Issue #12, opened 2018-04-10
Step 2E should include listbox with combobox


joanmarie:
(Spinoff issue from https://github.com/WhatSock/w3c-alternative-text-computation/issues/8)

AccName spec's step 2E states, "If the embedded control has role combobox, return the text alternative of the chosen option." Listbox is not mentioned here. At least some of us think that should be changed, with listbox and combobox being treated in the same fashion.

accdc:
Hi Matt,
The executable AccName prototype already supports all of the supported children of role=combobox, including Listbox, Grid, and Tree, such as the following tests:

```
<input type="text" id="test" />
  <label for="test">Flash the screen 
    <div role="combobox">
      <div role="textbox"></div>
            <div role="grid">
      <ul role="row" style="list-style-type: none;">
        <li role="gridcell" aria-selected="true">1</li>
   	<li role="gridcell">2</li>
   	<li role="gridcell">3</li>
      </ul>
    </div>
    </div>
    times.
  </label>
```

And

```

<input type="text" id="test" />
  <label for="test">Flash the screen 
    <div role="combobox">
      <div role="textbox"></div>
      <ul role="tree" style="list-style-type: none;">
        <li role="treeitem" aria-selected="true">1</li>
   	<li role="treeitem">2</li>
   	<li role="treeitem">3</li>
      </ul>
    </div>
    times.
  </label>
```

All of which are treated as widgets that support explicit values in accordance with the aria-selected model for selectable children.

I'm not sure if these are clearly conveyed in the spec though.

An ARIA Combobox only has a value when one of it's supported children like those above have aria-selected="true", or if the embedded textbox has a value (whether element.value if a native edit field or name from content if a simulated role=textbox), but if it has no value then nothing is returned as the value.

This is how I coded it into the AccName Prototype, but if others have a different interpretation of this, please let me know.

joanmarie:
Another thing which we need to address: What about multi-select listboxes which multiple items selected? 

(Sorry for not thinking about this until now. It's an argument for seeking/doing implementations early on in the process as per our new work flow.)

eps1lon:
It does now include `listbox`:
> If the embedded control has role combobox or listbox, return the text alternative of the chosen option.

But the point remains about `aria-multiselectable`. I guess

```diff
- [...] return the text alternative of the chosen option.
+ [...] return the text alternative of the selected options.
```
Would be sufficient and make it more obvious what "chosen" means by using the same terminology (`selected` and `aria-selected`).

PR #13, opened 2018-04-10
Treat the listbox role in the same fashion as combobox in step 2E

joanmarie:
Fixes github issue #12.


mcking65:
The change in this PR looks good to me. I have related questions that I will raise in issue #12.

jongund:
I agree with Matt, they should behave the same.

Jon

PR #53, opened 2019-06-18
Update the algorithm to handle roles where naming is not supported

joanmarie:
Add a statement to the end of Step 1 ("initialize") stating that if the
root node's role doesn't support naming, return the empty string. At the
present time, this impacts nothing because all roles currently support
naming. If that changes within the ARIA spec, this modification is meant
to accomplish the following:

`<span aria-label="bar">foo</span>`: The span has no name.

`<button><span aria-label="bar">foo</span></button>`: The button's
accessible name is "bar" (but the span itself has no name).

In other words, it prohibits generic elements themselves from getting
names, but preserves the name calculation results authors might be
expecting in currently-deployed content.

### DIFF
--- a/index.html
+++ b/index.html
@@ -249,7 +249,7 @@
     </section>
       <p>The text alternative for a given element is computed as follows:</p>
       <ol>
-        <li id="step1">Initialize:  Set the <code>root node</code> to the given element, the <code>current node</code> to the <code>root node</code>, and the <code>total accumulated text</code> to the empty string ("").</li>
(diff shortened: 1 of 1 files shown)


joanmarie:
@accdc and @jnurthen: Here is how I think we can keep the AccName spec in harmony with the ARIA spec even with names being prohibited (or unsupported) on generic roles. Please review. Thanks in advance!

carmacleod:
Pretty simple.  :)

JAWS-test:
> <button><span aria-label="bar">foo</span></button>: The button's accessible name is "bar" (but the span itself has no name).

If I understand the specification correctly, it works recursively, i.e. if a node contains child nodes, the name of the node is composed of the names of the child nodes. That's why it seems strange to me that a button containing a child node, where aria-label is ignored, has the name of the aria-label and not the text content as name

Issue #68, opened 2020-01-14
Support ARIA 1.3's aria-description property


aleventhal:
The precedence order is:
aria-describedby
aria-description

Pull request for aria-description in ARIA spec:
w3c/aria#1137

pkra:
I believe this was resolved by https://github.com/w3c/accname/pull/69 (just not automatically closed).

PR #69, opened 2020-01-14
Support aria-description

aleventhal:
Closes issue #68  and issue #77

Testing whether editing this comment triggers pr-preview.
Yay! It does! 🎉 

Editing comment again now to add `#mapping_additional_nd_description` in-page anchor for preview.

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/69.html#mapping_additional_nd_description" title="Last updated on May 16, 2021, 3:25 PM UTC (7335d41)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/69/d10fe80...7335d41.html" title="Last updated on May 16, 2021, 3:25 PM UTC (7335d41)">Diff</a>

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/69.html#mapping_additional_nd_description" title="Last updated on Nov 15, 2022, 6:18 PM UTC (a438283)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/69/f4eb23e...a438283.html" title="Last updated on Nov 15, 2022, 6:18 PM UTC (a438283)">Diff</a>


index.html
```diff
@@ -197,6 +197,7 @@ <h3>Name Computation</h3>
   <section id="mapping_additional_nd_description">
     <h3>Description Computation</h3>
     <p>If <a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a> is present, <a class="termref">user agents</a> MUST compute the accessible description by concatenating the text alternatives for elements referenced by an <code>aria-describedby</code> attribute on the current element. The text alternatives for the referenced elements are computed using a number of methods, outlined below in the section titled <a href="#mapping_additional_nd_te">Accessible Name and Description Computation</a>. </p>
+    <p>However, if <code>aria-describedby</code> is NOT present, and <a class="property-reference" href="#aria-description"><code>aria-description</code></a> IS present, <a class="termref">user agents</a> MUST use attribute value of <code>aria-description</code> as the accessible description.
```
carmacleod:
```suggestion
    <p>However, if <code>aria-describedby</code> is NOT present, and <a class="property-reference" href="#aria-description"><code>aria-description</code></a> IS present, <a class="termref">user agents</a> MUST use the attribute value of <code>aria-description</code> as the accessible description.
```

aleventhal:
@joanmarie what are the next steps for this one?

joanmarie:
> @joanmarie what are the next steps for this one?

Group review, including from @accdc who is taking the lead on this spec.

index.html
```diff
@@ -197,6 +197,7 @@ <h3>Name Computation</h3>
   <section id="mapping_additional_nd_description">
     <h3>Description Computation</h3>
     <p>If <a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a> is present, <a class="termref">user agents</a> MUST compute the accessible description by concatenating the text alternatives for elements referenced by an <code>aria-describedby</code> attribute on the current element. The text alternatives for the referenced elements are computed using a number of methods, outlined below in the section titled <a href="#mapping_additional_nd_te">Accessible Name and Description Computation</a>. </p>
+    <p>However, if <code>aria-describedby</code> is NOT present, and <a class="property-reference" href="#aria-description"><code>aria-description</code></a> IS present, <a class="termref">user agents</a> MUST use the attribute value of <code>aria-description</code> as the accessible description.
```
joanmarie: Nit: I wouldn't capitalize "NOT" or "IS."

aleventhal: Done

aleventhal:
Hi @joanmarie, there should not be a change to the algorithm. When aria-describedby is used, it uses the name algorithm as it stands now to get the text. That algorithm does not consider description markup, it only returns the primary text.

When aria-describedby is not used, then aria-description can be used to set the description. However, aria-description does not affect the recursive algorithm, whether it's used for name or description.

joanmarie:
Ok.... Though I still think it's odd that the algorithm for computing the name and description lacks any mention of `aria-description`. If no one else thinks so.... :woman_shrugging:

aleventhal:
> Ok.... Though I still think it's odd that the algorithm for computing the name and description lacks any mention of `aria-description`. If no one else thinks so....

As it stands, we have these 3 headings, which is confusing already:
Accessible Name Computation
Accessible Description Computation (which mentions aria-description)
Accessible Name and Description Computation (reused by the others)

IMO the overlap between these names is kind of confusing already. The last one could be renamed "Primary Text for a Subtree Computation". Then the first two would explain that the Primary Text for a Subtree is used for aria-labelledby and aria-describedby.

But it doesn't make sense to mention aria-description in that Primary Text Computation, because that would change the way aria-description works, which is not desired.

JAWS-test:
See also https://github.com/w3c/accname/issues/77

carmacleod:
@aleventhal Notes from the meeting bot:
> aaronlev: I like that name. "Text Content Computation"
> aaronlev: how about if I just expand the current pr to add this?

There was general agreement (not in the notes because the scribe was lagging behind... 😄 ).

aleventhal:
I think this is is a step in the right direction, but doesn't solve everything.

In the table, I mention all relevant HTML markup, because currently I think the spec dances around HTML markup, yet seems to desperately want to mention it.  HTML should be treated as a first class citizen by this spec, maybe even SVG too. Otherwise it's hard to define an order of markup precedence, let alone which markup should be processed. Getting names and descriptions right in HTML is too important to just hand wave around this.

accdc:
Hi, I think this is a great step in the right direction. Please merge if you will. We will continue working on the rest as well after.

JAWS-test:
@aleventhal, many thanks for the comprehensive revision. I think it removes all the ambiguities and errors that were previously contained. In the preview, there is only one problem with the numbering of the list items: instead of letters, numbers are used, so that the references (e.g. "2F.i") no longer work. But maybe this is just a CSS problem of the preview.

The table for description is very good. I just think the first column header is wrong (it should be "precedence" instead of "predence", right?)

aleventhal:
@accdc I'm not authorized to merge, unfortunately. Do you want to authorize me or merge?

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+          <th>Applicable conditions</th>
+          <th>How used to compute description</th>
+        </tr>
+      </thead>
+      <tbody>
+        <tr>
+          <th>1</th>
+          <td><a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a> attribute</td>
+          <td>Use on any element</td>
+          <td><a href="#mapping_additional_nd_te">Text content computation</a> on any relevant nodes, concatenated, and separated by a space character</td>
```
carmacleod: "any relevant nodes" seems a little vague.
```suggestion
          <td><a href="#mapping_additional_nd_te">Text content computation</a> on all nodes referenced by aria-describedby on the element, concatenated, and separated by a space character</td>
```

aleventhal: That's fine. We may as well be as exact as possible where it's not too verbose. In general, I think we should work to make this spec more concise and readable for browser developers.

carmacleod:
Please don't merge yet - we need to run through the approvals. I had previously given approval, but this is a significant change, so I needed to look it over again. I love the new section title - that helps a lot. And personally, I like the new table with the HTML and I think it's helpful (obviously a similar table would have to be created for accessible name as well), **but please note** that it duplicates some content from the [HTML-ARIA spec](https://w3c.github.io/html-aam/) (look for instances of "Accessible Description Computation"). As such, I think we all need to decide whether or not this is definitely the way we want to go, because it may be problematic for maintenance. An alternative is to link from the table into the HTML-ARIA spec (not as nice or convenient for authors, but doable). Another alternative is for the HTML-ARIA spec to link into ACCNAME (which it already does in other places, so maybe not too crazy).

I assume that @accdc has approved, but @joanmarie and I still need to approve, and I am adding @scottaohara to the list of reviewers because I think it's really important that this spec duplication is noted and discussed.

aleventhal:
No problem @carmacleod ... I didn't know what the process was for this spec.

Is this spec intended for browser devs only, or also web authors?

aleventhal:
Regarding creating a table for name:
- Unlike description, name should be able to directly use the results of the Text Content Computation
- We'll want to have 1 or more tables in the Text Content Computation
- I don't think we should hold this PR up for that. I think we're going to find that we want to make a large number of clarification, readability and correctness changes and it will be really hard to move forward if we try to do them all at once.

carmacleod:
Sorry, forgot to answer this question:
> Is this spec intended for browser devs only, or also web authors?

The [first sentence](https://w3c.github.io/accname/#abstract) says:
> This document describes how user agents determine the names and descriptions of accessible objects from web content languages.

and there's no mention of authors other than in the context of "author intent" or "provided by the author".

Also, I don't want to hold up this PR until AllTheThings are done either. I just want to make sure that we are all ok with having accname take on the added responsibility of defining (or keeping up with) the specifics of HTML accessible descriptions (and possibly accessible names as well) that are already specified in HTML-ARIA.

aleventhal:
> > Unlike description, name should be able to directly use the results of the Text Content Computation

> Nope. Look for instances of "Accessible Name Computation" ....

Thanks @carmacleod , what I meant by this, is that the HTML that applies to names would go into the Text Content Computation and would feed both the Name Computation and Description Computation that way. 

In a way the algorithm works like this:
- Name : directly use Text Content Computation
- Description : if aria-describedby, use Text Content Computation, otherwise use the other markup listed in the table (4.2)
- Native markup, aria-label, etc. all feed the Text Content Computation, which then feed the name and description computations

Right now the Text Content Computation does some handwaving around native semantics that apply to names. It desperately wants to tell us about alt text, placeholder, etc. Meanwhile us browser developers really just desperately want to know what goes into it without needing to consult a bunch of different documents.

I'm more worried about making this document clear & useful and helping browser devs do the right thing than about which doc should or shouldn't have responsibility for normative requirements.

carmacleod:
> Right now the Text Content Computation does some handwaving around native semantics that apply to names. It desperately wants to tell us about alt text, placeholder, etc. Meanwhile us browser developers really just desperately want to know what goes into it without needing to consult a bunch of different documents.

> I'm more worried about making this document clear & useful and helping browser devs do the right thing than about which doc should or shouldn't have responsibility for normative requirements.

Completely understood. And it would be great to replace all handwaving with tables that have real content. I agree that it would make this document much more clear & useful.

Still want to have buy-in from all concerned parties for the new direction.

accdc:
Hi, it's important to note that this is in regard to the core AccName spec, and as such, is not primarily intended for web authors, but for user agents. Rather, this is how platforms should be building the accessibility tree and computing the accessible name and description within it. Since it is not HTML specific, it isn't actually dependent on the ARIA-HTML spec for anything, but is actually the other way around. Derived naming specs should be using the core AccName algorithm to shape their own algorithms within their native host semantics as needed, but their individual requirements shouldn't prevent the core AccName spec from moving forward.

aleventhal:
Talking to James Teh, and internally both Chrome and Firefox call the recursive algorithm the Text Equivalent Computation. I'd like to rename it to that to match what we actually call it.
Hopefuly everyone is ok with that.

jcsteh:
To provide some more background on my conversation with @aleventhal, I have major concerns about the name "Text Content Computation". "Name" and "text content" are a11y concepts that are already conflated by many people, especially web authors. They are not the same, and in fact, "name" (accessible name property; e.g. IAccessible::accName) can be different to "text content" (as returned by a11y text interfaces; e.g. IAccessibleText). I cover this difference and the problems it causes at length in [a blog post](https://www.jantrid.net/2015/12/woe-aria-surprisingly-but-ridiculously.html).

Given these concerns and the fact that both Gecko and Chromium internally call this "Text Equivalent Computation", we figured this was a less "loaded" name.

JAWS-test:
if Accname should not rely too much on HTML and avoid redundancies (with HTML AAM), then I would suggest the following:

The table in chapter [4.2](https://pr-preview.s3.amazonaws.com/w3c/accname/pull/69.html#mapping_additional_nd_description) can be shortened to the following lines:
* aria-describedby
* aria-description
* special forms of host language description (like value, caption, summary, placeholder etc.) - with a link to HTML AAM (summarizing this would also be correct in terms of the precedence)
* Tooltip (title)

JAWS-test:
> I like the new table with the HTML and I think it's helpful (obviously a similar table would have to be created for accessible name as well

I also think the table is very good and suggest that such a table is also created for Name in chapter 4.1 with the following lines:
* aria-labelledby
* aria-label
* Label methods of the host language (such as label, caption etc.; with link to HTML AAM)
* Text content (including CSS content)
* Tooltip

The table would only need to indicate that it is a simplified representation. Detailed information is provided in Chapter 4.3, especially on special cases such as
* hidden contents (step 2.A)
* traversal (step 2.B)
* recursion (step 1 and following)
* nested form fields (step 2.E)

carmacleod:
I spoke with @scottaohara offline, and he is ok with this, so I am ok with it, too.

He says, "html aam is supposed to be under the ARIA umbrella now….so i don’t mind the duplication too much if i can just point to other specs to reinforce the one i work on".

So I think we're just waiting for @joanmarie to review.

joanmarie:
I dig the table. Would it be a lot of work to do the same for name like @JAWS-test suggested?

aleventhal:
@joanmarie I assume you mean in a different PR? I'd be willing to help with that, but will want to agree on a general approach. I think we should file a separate issue for it. See my comment above (https://github.com/w3c/accname/pull/69#issuecomment-594728866) for how I imagine it being done, but others have slightly different ideas, so I think we should hash that out.

index.html
```diff
@@ -3,7 +3,7 @@
 <head>
 <title>Accessible Name and Description Computation 1.2</title>
 <meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
-<script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove"></script>
+<script src="https://www.w3.org/Tools/respec/respec-w3c-common" class="remove" defer></script>
```
cookiecrook: If this is needed, it should be in a separate pull request with explanation.

Also, I haven't looked at ReSpec behavior in the last few years, but since it's used to render a lot of the spec contents, I'm not sure you'd want to defer its load. It might trigger the dreaded FOUC: "flash of unstyled/untransformed content."

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+    <h3>Text Equivalent Computation</h3>
+    <p>The text equivalent computation is used to by both the <a class="termref">accessible name</a> and <a class="termref">accessible description</a>. There are different rules provided for several different types of <a class="termref">elements</a>, <a class="termref" data-lt="node">nodes</a>, and combinations of markup. Text alternatives are built up, when appropriate, from all the relevant content contained within an <a class="termref">element</a>. This is accomplished via steps 2B and 2F, which are recursive, using the full set of rules to retrieve text from its own children or nodes it references. </p>
```
cookiecrook: Typo? "used to by both"

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+          <td><a class="property-reference" href="#aria-describedby"><code>aria-description</code></a> attribute</td>
+          <td>Use on any element</td>
+          <td>As a flat string</td>
+        </tr>
+        <tr>
+          <th>3</th>
+          <td>HTML <code>value</code> attribute</td>
+          <td>
+            <ul>
+              <li>Only use on HTML <code>&lt;input type="button"|"submit"|"reset"&gt;</code></li>
```
cookiecrook: "or equivalent host lang elements"?

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+        </tr>
+        <tr>
+          <th>2</th>
+          <td><a class="property-reference" href="#aria-describedby"><code>aria-description</code></a> attribute</td>
+          <td>Use on any element</td>
+          <td>As a flat string</td>
+        </tr>
+        <tr>
+          <th>3</th>
+          <td>HTML <code>value</code> attribute</td>
```
cookiecrook: "or equivalent host language attribute"?

index.html
cookiecrook: Possibly something here to address a similar concept in web components? Maybe even phrased generically so it can address AOM virtual nodes if/when those become standard?

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+            </ul>
+          </td>
+          <td><a href="#mapping_additional_nd_te">Text equivalent computation</a> on the summary element</td>
+        </tr>
+        <tr>
+          <th>6</th>
+          <td>HTML <code>title</code> attribute</td>
+          <td>
+            <ul>
+              <li>Any HTML element</li>
```
cookiecrook: any element?

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+            <ul>
+              <li>Only use for a corresponding HTML <code>&lt;details&gt;</code> parent, if the first child of the <code>&lt;details&gt;</code></li>
+              <li>Only use if not already used for the accessible name of the corresponding <code>&lt;details&gt;</code></li>
+            </ul>
+          </td>
+          <td><a href="#mapping_additional_nd_te">Text equivalent computation</a> on the summary element</td>
+        </tr>
+        <tr>
+          <th>6</th>
+          <td>HTML <code>title</code> attribute</td>
```
cookiecrook: host language tooltip attribute or equivalent concept (e.g. HTML `title` attr)

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+            <ul>
+              <li>Only use on HTML <code>&lt;input type="button"|"submit"|"reset"&gt;</code></li>
+              <li>Only use if not already used for the accessible name of this node</li>
+            </ul>
+          </td>
+          <td>As a flat string</td>
+        </tr>
+        <tr>
+          <th>4</th>
+          <td>HTML <code>&lt;caption&gt;</code> child of a <code>&lt;table&gt;</code></td>
```
cookiecrook: Isn't `caption` already covered as a ~"host language labeling mechanism" for `table`?

Regardless, I think it's okay to use `html:caption` as the only known example instance of a more specific rule, but the rule should not limit itself to HTML.

index.html
```diff
@@ -187,20 +187,89 @@ <h2>Name and Description</h2>
...
+          <td>
+            <ul>
+              <li>Only use for a corresponding HTML <code>&lt;table&gt;</code> parent, if the first child of the <code>&lt;table&gt;</code></li>
```
cookiecrook: 1. Remove HTML-specific language.
2. If this isn't already covered by host lang label mech, the "first child" reference here does not account for caption position‡, which may be relevant. perhaps, ~"first descendant rendered to the user"? We'd need to define this, such as left to top caption in LTR languages like English, and right or top caption in RTL langs like Arabic.

‡ caption position can still be defined by the deprecated HTML `align` attribute, or the non-deprecated `caption-side` CSS property. left/right values for the latter have been deprecated though, so maybe it's sufficient to reference top/bottom placement directly?

cookiecrook:
ARIA 1.0 called this "Text Alternative Computation"
https://www.w3.org/TR/wai-aria-1.0/roles#textalternativecomputation

I don't have a preference for "alt" or "equiv", just an FYI. 

screen shot of 3 headings from ARIA 1.0 archive, enumerated in the prose below this image

5.2.7.1. Name Computation
5.2.7.2. Description Computation
5.2.7.3. Text Alternative Computation

jcsteh:
I'm okay with Text Alternative Computation in case that's the consensus.
That said, the fact that both Firefox andCromium (not sure about WebKit)
call this Text Equivalent Computation leads me to prefer that very slightly.

joanmarie:
Related question: What happens if the author explicitly sets `aria-description` to an empty string in order to prevent an HTML `title` attribute from being presented by a screen reader?

aleventhal:
> Related question: What happens if the author explicitly sets `aria-description` to an empty string in order to prevent an HTML `title` attribute from being presented by a screen reader?

 It will end up as an empty description. The condition is covered by this:

. <a class="termref">User agents</a> MUST use the first applicable entry from the table where the listed conditions are met, as described in the last column. The user agent MUST NOT use any markup other that the first relevant markup found, even if that markup results in an empty description:</p>

joanmarie:
Thanks Aaron. Best place to hide things is right in front of me.... :-/

scottaohara:
`aria-label` being set to the empty string is supposed to be ignored from my understanding of the algorithm.

i can understand the rational behind why that should be ignored (avoid instances of no accessible name).

but per that rational i would expect `aria-description=""` to also be ignored and not create an empty description to be consistent.

I understand that preventing an empty name is more pressing than preventing an empty description, but being that ARIA is already confusing enough for authors, I'd question the divergence here even though I can respect the potential reasoning behind it.

aleventhal:
@scottaohara , good point. We should check how that actually works across browsers. IMO aria-label="" should mean there is no name, unless there is markup that takes precedence (aria-labelledby).

I always thought that aria-label="" wouldn't make sense unless you were trying to block a name, and I believe I've coded to match that expectation. It's possible that the spec, which was written to try to match what browsers do (rather than the other way around), should be updated.

scottaohara:
in the ARIA spec (5.2.8 Accessible Name Calculation, 3rd list item):

>prohibited: the element has no name. Authors MUST NOT use the aria-label or aria-labelledby attributes to name the element.

I'm a bit on the fence about going back on that.  So I'll let others discuss to help me determine if i should be more hard line or lax on that point :)

accdc:
Hi,
Before I merge this, I wanted to check first, is the WG in agreement for these changes? 

Since this is already implemented within browsers, it seems like a good idea to pull this in to match spec and implementation.

If anyone objects, please say so sooner than later.

index.html
MelSumner, line 272:
```suggestion
    <p>The text equivalent computation is used by both the <a class="termref">accessible name</a> and <a class="termref">accessible description</a>. There are different rules provided for several different types of <a class="termref">elements</a>, <a class="termref" data-lt="node">nodes</a>, and combinations of markup. Text alternatives are built up, when appropriate, from all the relevant content contained within an <a class="termref">element</a>. This is accomplished via steps 2B and 2F, which are recursive, using the full set of rules to retrieve text from its own children or nodes it references. </p>
```

MelSumner:
@aleventhal if you can give some feedback on some of the open comments, I can help get this over the finish line. 👍

aleventhal:
@MelSumner Are you specifically talking about @cookiecrook 's comments? They all make sense. I think he just wants to future-proof it by using spec text that doesn't prevent mappings of other markup languages besides HTML.

Are you ok with applying his suggestions?

accdc:
Hi, I've rereviewed all of the thread, which I apologize took longer for me to make sense than I expected, and I think with the changes proposed by @cookiecrook  that adding this would be a good idea. Unfortunately I am having a lot of trouble untangling the ball of yarn that this has become, so I would very much appreciate a new PR if possible that simply incorporates these changes in a way that we can merge it. @aleventhal  or @cookiecrook  can you help with this?

@scottaohara I understand your concerns about diverging from what occurs when aria-label is set to an empty string, but there is value in suppressing the description at times. I have an example of this that is already using this technique at:
http://whatsock.com/Templates/Sliders/Vertical/index.htm

The issue I was having is that the slider, when it received focus, was having the computed difference between the min and max values spoken as a description, even though this was not something I wanted because it uses a custom value string as the value. The solution was to add aria-description=" " to the slider, which suppressed this from occurring. E.G. if you set focus to the slider with JAWS running and change to another value using the Up/Down arrow keys, then press Insert+Tab to say the current control name and description, you will not hear the unwanted information. (Tested with JAWS2022 in Chrome Beta).

Are you okay with this type of functionality in edge cases such as this?

index.html
cookiecrook:
```suggestion
          <td>HTML <code>value</code> attribute, or equivalent host language attribute</td>
```

index.html
cookiecrook:
```suggestion
              <li>Only use on HTML <code>&lt;input type="button"|"submit"|"reset"&gt;</code> or equivalent host language features</li>
```

index.html
cookiecrook:
```suggestion
          <td>HTML <code>&lt;caption&gt;</code> child of a <code>&lt;table&gt;</code> or equivalent host lang feature</td>
```

index.html
cookiecrook:
```suggestion
          <td>host language tooltip attribute or equivalent feature (e.g. HTML <code>title</code> attribute)</td>
```

index.html
cookiecrook:
```suggestion
              <li>Any element</li>
```

cookiecrook (commented):
I reposted most comments (all save one) as diffs. Will leave these for a few days for comment and try to propose wording for the last comment on table captions. If there are no objections or comments, we can commit these into the prior PR. I think that will resolve @accdc's request to update the PR.

cookiecrook:
Only the very last diff came through in the review comment above... Check the [Filed Changed](https://github.com/w3c/accname/pull/69/files) view for the other +/- diffs.

index.html
cookiecrook: I need to come back to this one for a diff change suggestion.

accdc:
@cookiecrook  Thank you!

scottaohara:
@accdc regarding "Are you okay with this type of functionality in edge cases such as this?"

practically i would say 'yes', i get why you'd want that / do that.  my worry though comes from teams that are using frameworks where they just specify aria attributes on their components, and set them to the 'false` or empty string values when they're not in use.  

so, while your example makes sense because you know what you're doing, there would be a number of developers who are implementing widely used framework components that are doing the same thing, but also specifying a `title` attribute, for example, intending for it to be the accessible description... but that won't be exposed if `aria-description=""` overwrites it.

so if `aria-description=" "` wins out and `aria-description=""` is ignored, then i think _that_ would be ok and would mitigate against the issues I know would arise otherwise.  that sound ok to you?

accdc:
"so if aria-description=" " wins out and aria-description="" is ignored, then i think that would be ok and would mitigate against the issues I know would arise otherwise. that sound ok to you?"

Absolutely. That is what Aaron originally proposed I believe, which I'm pretty sure is already baked into Webkit  at this time.

scottaohara:
great.  thanks for confirming, Bryan

index.html
cookiecrook: Coming back to this one, maybe it's best to remove `html:caption` altogether, since it should already be covered by ~"host language labeling mechanism" elsewhere in the computation. Thoughts?

cookiecrook (changes requested):

accdc:
"Coming back to this one, maybe it's best to remove html:caption altogether, since it should already be covered by ~"host language labeling mechanism" elsewhere in the computation. Thoughts?"

That sounds good to me.

scottaohara:
So after giving this a careful review and per James Craig's comments and the [comment jaws test made](https://github.com/w3c/accname/pull/69#issuecomment-595017008), I do think that it probably makes sense to either simplify the specific HTML features into a single row to reference HTML AAM and the elements that have unique features that would slot into this part of the computation, or we'd need to expand on these sections to include more / potentially make some adjustments and corrections.  I tend to lean on the former rather than the latter.

scottaohara:
So after giving this a careful review and per James Craig's comments, I do think that it probably makes sense to either simplify the specific HTML features into a single row to reference HTML AAM and the elements that have unique features that would slot into this part of the computation, or we'd need to expand on these sections to include more / potentially make some adjustments and corrections.

accdc:
@scottaohara I'm actually confused as to what is being recommended at this point. Do you vote against the PR until further issues are addressed? Other than the removal of the table captions content, I'm not clear on what issues remain for this PR.
Thanks,
Bryan

scottaohara:
assigning myself to take on the task of revising HTML AAM elements description algorithms to follow this format and then update this PR to point to the elements that have variants - essentially replacing rows 3 - 5.

cookiecrook:
My review was re-requested on this PR, but none of my previous proposed changes from March 2022 were committed. I can commit them to the PR myself if the other editors are okay with that, but I'd prefer another editor do it. Once that's done, I can review again.

accdc:
I'm never going to be as good with github as others here, so if you or Melanie could help with this I'd appreciate it.

cookiecrook:
From call: Dependent on #433? @scottaohara

index.html
```diff
@@ -200,22 +200,91 @@ <h2>Name and Description</h2>
...
+      <tbody>
+        <tr>
+          <th>1</th>
+          <td><a class="property-reference" href="#aria-describedby"><code>aria-describedby</code></a> attribute</td>
+          <td>Use on any element</td>
+          <td><a href="#mapping_additional_nd_te">Text content computation</a> on all nodes referenced by aria-describedby on the element, concatenated, and separated by a space character</td>
+        </tr>
+        <tr>
+          <th>2</th>
+          <td><a class="property-reference" href="#aria-describedby"><code>aria-description</code></a> attribute</td>
```
cookiecrook, line 234:
```suggestion
          <td><a class="property-reference" href="#aria-description"><code>aria-description</code></a> attribute</td>
```

mcking65:
Do we intend the aria-description value on an element to be able to be included in its description by a self-reference in the value of aria-describedby?

If so, that does not appear to be supported by the current definition of aria-describedby computation.

If not, we may want to make that constraint more clear. It would make the behavior of aria-describedby slightly different from the behavior of aria-labelledby.

JAWS-test:
> Do we intend the aria-description value on an element to be able to be included in its description by a self-reference in the value of aria-describedby?

I don't think that makes sense. If elements are referenced by aria-describedby, then their AccName is relevant and not their AccDesc. I.e. the AccName of an element becomes the AccDesc of the element which is referred to by aria-describedby. The AccDesc of an element does not matter (whether via title, aria-description or any other method) if the element is referenced via aria-describedby.

index.html
```diff
@@ -200,22 +200,70 @@ <h2>Name and Description</h2>
...
+					<td>
+						Unique host language features MAY participate in the description computation for an element, only if they were not already used
+						for the accessible name of the applicable element. See <a data-cite="html-aam/#accdesc-computation">HTML AAM: Accessible Description Computation</a> 
+						for the HTML elements which meet this condition.
+					</td>
+					<td>
+						Either a <a href="#mapping_additional_nd_te">text equivalent computation</a> of the element, or as a flat string of the attribute's value.
+					</td>
+				</tr>
+        <tr>
```
cookiecrook, line 250:
```suggestion
        </tr>
        <tr>
          <th>3</th>
          <td>host language features which participate in the description calculation</td>
          <td>
            Unique host language features MAY participate in the description computation for an element, only if they were not already used
            for the accessible name of the applicable element. See <a data-cite="html-aam/#accdesc-computation">HTML AAM: Accessible Description Computation</a> 
            for the HTML elements which meet this condition.
          </td>
          <td>
            Either a <a href="#mapping_additional_nd_te">text equivalent computation</a> of the element, or as a flat string of the attribute's value.
          </td>
        </tr>
        <tr>
```

Nit. Match the surrounding whitespace indentation.

cookiecrook (approved):

PR #84, opened 2020-06-17
Adding some exports to help with xref/Respec

aarongustafson:


aarongustafson:
These are changes we are requesting to unblock https://github.com/w3c/image-resource/pull/28

marcoscaceres:
This may be exporting too much. We should only export the things we immediately need (i.e., only "accessible name", I think).

jnurthen:
Changes to things in the common directory need to be made against the aria-common repository.

________________________________

aarongustafson:
Thanks @jnurthen!

marcoscaceres:
FWIW, accessible name should still be defined in this specification, and not included in the common set (the common set is not good idea, because it means a bunch of specs are redefining things over and over again and no knows which spec to cite for a term - or which spec should export which term).

jnurthen:
@marcoscaceres FWIW I agree with you - I just don't have the time to action it right now.

aarongustafson:
There is a link to the term in this spec. Would wrapping that in a dfn and expiring that work? I'm happy to make that change & resubmit.

jnurthen:
All the terms are defined in aria-common right now. I believe we need to go through all the terms and move them to the relevant spec. I don't think we can 1-off a single term in this way.

jnurthen:
Accessible Name is now exported from the ARIA spec. 
Including  "wai-aria" in the xref will pick up the definition.

Issue #36, opened 2018-11-06
Step D needs clarification


john-cotter:
Why is [title] mentioned in step D? Title should not be considered until the end of the calculation (step I).

If title were considered at step D for all html elements, that would put title before a name-from-content element's subtree (step F).

Example:
`<button title="there">hello</button>`
As currently written, the algorithm would produce an accName/accDesc of "there hello".
Real world screen reader testing produces "hello there".

Should title only be used in step D for specific HTML elements, such as an iframe? If so clarification is needed as to what those elements are.

Furthermore, in step D, it would be helpful to have a significant amount of clarification for all native html elements.  For example, when a form element has an associated label, should it only grab the innerText of the label, or should the algorithm be treating the label like a name from content element, recursively traversing through the label's child nodes? Also, should the algorithm consider whether the label has an aria-label or title attribute?

carmacleod:
The [HTML AAM spec](https://w3c.github.io/html-aam/#accessible-name-and-description-computation) contains specific AccName info for native html elements.

toddbri:
I have a similar question regarding Rule D. As I read the current TAN computation the following markup should return ‘Title Text’ as the accName for h1.

Example:

```
<h1 title="Title Text">
      <img src="images/destination.com/image.png" alt="Image Alt">
       <div>
            <div>Part 1</div>
            <strong>Part 2</strong>
        </div>
</h1>
```
Both Chrome and Safari are returning the accName as ‘Image Alt Part 1 Part2” in their respective accessibility trees.

What is the correct calculation of the accName for h1?

jnurthen:
See also https://github.com/w3c/accname/issues/10

accdc:
Hi,
In the case of your H1 code, the accessible name would only be the accumulated names from the inner content, and not the title.

So, accName = “Image Alt Part 1 Part 2”, and accDescription = “”.

The reason being that, the word title in the spec in this context wasn't meant to refer to the title attribute, but rather that of alt text, as noted in issue 10. I'll work on getting that fixed for 1.2 soon.

For future reference, there is a mirror project on GitHub that provides a live executable representation of what we expect the accessible name and description computation to be based on what the AccName spec is meant to replicate within browsers, at
https://github.com/WhatSock/w3c-alternative-text-computation

[...]

PR #88, opened 2020-08-13
Added HTML-AAM reference to step 2E.

OisinNolan:
- Added a link to the HTML-AAM to step D.

The [HTML-AAM](https://www.w3.org/TR/html-aam/) is relevant to accessible name computation for HTML (step 2D in particular) but is not linked anywhere in the spec. This was brought up in [issue #36](https://github.com/w3c/accname/issues/36).

***
<a href="https://pr-preview.s3.amazonaws.com/OisinNolan/accname/pull/88.html" title="Last updated on May 16, 2021, 3:14 PM UTC (45e5f56)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/88/9e7c441...OisinNolan:45e5f56.html" title="Last updated on May 16, 2021, 3:14 PM UTC (45e5f56)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -308,7 +308,8 @@
 &lt;/ul&gt;</code></pre>
                 </details></div>
             </li>
-            <li id="step2D">Otherwise, if the <code>current node</code>'s native markup provides an <a class="termref">attribute</a> (e.g. <code>title</code>) or <a class="termref">element</a> (e.g. HTML <code>label</code>) that defines a text alternative, return that alternative in the form of a <code>flat string</code> as defined by the host language, unless the element is marked as presentational (<code>role="presentation"</code> or <code>role="none"</code>). See <a href="https://www.w3.org/TR/html-aam/">HTML-AAM</a> for detail on HTML markup that defines a text alternative.
+            <li id="step2D">Otherwise, if the <code>current node</code>'s native markup provides an <a class="termref">attribute</a> (e.g. <code>title</code>) or <a class="termref">element</a> (e.g. HTML <code>label</code>) that defines a text alternative, return that alternative in the form of a <code>flat string</code> as defined by the host language, unless the element is marked as presentational (<code>role="presentation"</code> or <code>role="none"</code>).
+              <div class="note">See <a href="https://www.w3.org/TR/html-aam-1.0/#accessible-name-and-description-computation">HTML-AAM</a>, <a href="https://www.w3.org/TR/svg-aam-1.0/#mapping_additional_nd">SVG-AAM</a>, or other host language documentation for more information on markup that defines a text alternative.</div>
               <div><details>
                 <summary>Comment:</summary>
                 <p>For example, in <abbr title="Hypertext Markup Language">HTML</abbr>, the <code>img</code> element's <code>alt</code> attribute defines a text alternative string, and the <code>label</code> element provides text for the referenced form element.  In <abbr title="SVG2">SVG2</abbr>, the <code>desc</code> and <code>title</code> elements provide a description of their parent element. </p>
--- a/index.html
+++ b/index.html
@@ -3,17 +3,18 @@
 <head>
 <title>Accessible Name and Description Computation 1.2</title>
 <meta content="text/html; charset=utf-8" http-equiv="Content-Type"/>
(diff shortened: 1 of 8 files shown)


index.html
```diff
@@ -308,7 +308,7 @@ <h4>Terminology</h4>
 &lt;/ul&gt;</code></pre>
                 </details></div>
             </li>
-            <li id="step2D">Otherwise, if the <code>current node</code>'s native markup provides an <a class="termref">attribute</a> (e.g. <code>title</code>) or <a class="termref">element</a> (e.g. HTML <code>label</code>) that defines a text alternative, return that alternative in the form of a <code>flat string</code> as defined by the host language, unless the element is marked as presentational (<code>role="presentation"</code> or <code>role="none"</code>).
+            <li id="step2D">Otherwise, if the <code>current node</code>'s native markup provides an <a class="termref">attribute</a> (e.g. <code>title</code>) or <a class="termref">element</a> (e.g. HTML <code>label</code>) that defines a text alternative, return that alternative in the form of a <code>flat string</code> as defined by the host language, unless the element is marked as presentational (<code>role="presentation"</code> or <code>role="none"</code>). See <a href="https://www.w3.org/TR/html-aam/">HTML-AAM</a> for detail on HTML markup that defines a text alternative.
```
eps1lon, line 311: Is there a specific section we can link to? Otherwise considering a full spec as "details" might be a bit overwhelming. Maybe
```suggestion
            <li id="step2D">Otherwise, if the <code>current node</code>'s native markup provides an <a class="termref">attribute</a> (e.g. <code>title</code>) or <a class="termref">element</a> (e.g. HTML <code>label</code>) that defines a text alternative, return that alternative in the form of a <code>flat string</code> as defined by the host language, unless the element is marked as presentational (<code>role="presentation"</code> or <code>role="none"</code>). See <a href="https://www.w3.org/TR/html-aam/#accessible-name-and-description-computation">HTML-AAM</a> for detail on HTML markup that defines a text alternative.
```

This was suggested by @carmacleod  in https://github.com/w3c/accname/issues/36#issuecomment-436278363

OisinNolan: I agree, that section seems most relevant.

cookiecrook:
The ARIA WG discussed on August 13th, 2020. Advice for the PR follows:

1. The current PR (specifically the language "See HTML-AAM") makes it seem like HTML is the only relevant reference. Since ARIA is intended to work in multiple host languages, including but not limited to HTML, the editor's would prefer this link at least two external references... "For example, see HTML-AAM, SVG-AAM, or other host language documentation for more information..." 
2. Also, any non-normative reference inside a normative should e delineated by a non-normative grouping. In this case, probably `<div class="note">Note: [your text here]</div>`

cookiecrook:
@OisinNolan see above... Either post an updated PR, or one of the editors can update the PR for you. Thanks.

cookiecrook:
@jnurthen @carmacleod There was another note in the minutes: "should target the title change in next version of AccName" of which I don't recall the meaning... (I blame the scribe). Any ideas?

carmacleod:
@cookiecrook 
> should target the title change in next version of AccName

The next comment in [the minutes](https://www.w3.org/2020/08/13-aria-minutes.html) says:
> jamesn: that's #36 that addresses the other AccName concern re: title

@jnurthen can correct me if I am wrong, but I think it's because there was some question about whether title should be removed from step D, in which case we wouldn't want to give it as an example. Can we just get around this completely by using `alt` as the example of "an attribute that defines a text alternative" in step D? i.e.

> Otherwise, if the current node's native markup provides an attribute (e.g. alt) or element (e.g. HTML label) that defines a text alternative, ...etc... followed by the Note that links to the 2 AAM specs.

scottaohara:
@carmacleod we mentioned `title` was appropriate here if it was in context to `<title>` of SVGs.

The concern about this being HTML's `title` attribute is that it is so low on the ladder for providing an accName, that there was a concern alluding to it here would make it seem more important in the accName calculation than it is / should be.

MelSumner (approved):
Resolved merge conflicts, and edited title to reflect current step order. Looks good to go.

accdc (approved):
Looks good to me.

PR #92, opened 2020-09-01
Removed redundant line from 2.C.

OisinNolan:
The second bullet point under step 2.E. states `If the embedded control has role menu button, return the text alternative of the button.`.

I don't see any reason to distinguish `menu button` from `button` here. I believe that if we are to remove the bullet point quoted above (as is done in this PR) then such `menu button`s would simply be handled by 2.F, given that nodes with role `button` allow name from content.

The motivation here is to simplify the spec by removing redundancy.

If I have missed something here please let me know, thanks! :)

***
<a href="https://pr-preview.s3.amazonaws.com/OisinNolan/accname/pull/92.html" title="Last updated on May 16, 2021, 3:04 PM UTC (49caffd)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/92/9e7c441...OisinNolan:49caffd.html" title="Last updated on May 16, 2021, 3:04 PM UTC (49caffd)">Diff</a>


MelSumner (approved):
This makes sense based on the information provided and seems like an editorial fix. If @accdc agrees we can merge.

accdc (approved):
Looks good to me.

PR #112, opened 2021-02-19
Only use accumulated text that is not empty

### LINKED ISSUE #72: aria-label and aria-labelledby are inconsistent when they point to a blank label
# aria-label
```
<h1 aria-label=''>Example Heading</h1>
```
the calculated acc-name is 'Example Heading' when aria-label is blank, because the algorithm ignores empty or blank aria-labels at step 2C. Browser results for this are:

* FF 72: 'Example Heading' 
* Safari 18: 'Example Heading' 
* Chrome 79: Name: 'Example Heading' 

# aria-labelledby
```
<span id="label"></span>
<h1 aria-labelledby="label">Example Heading</h1>
```
the calculated acc-name is '' when aria-labelledby points to a blank element because the algorithm returns at 2B iii) as soon as it finds a valid aria-labelledby reference - even if the target is blank. Browser results for this are:

* FF 72: 'Example Heading' 
* Safari 18: 'Example Heading' 
* Chrome 79: ''

Given the majority of implementations ignore an aria-labelledby if it points to an empty element, and the results are better that way, it might be better aligning the spec with this.

kasperisager:
This seemed like the simplest way to resolve #72.

***
<a href="https://pr-preview.s3.amazonaws.com/kasperisager/accname/pull/112.html" title="Last updated on Feb 19, 2021, 3:07 PM UTC (fe3a05d)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/112/2399041...kasperisager:fe3a05d.html" title="Last updated on Feb 19, 2021, 3:07 PM UTC (fe3a05d)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -271,7 +271,7 @@
                         <li id="step2B.ii.c">Append the <code>result</code>, with a space, to the <code>accumulated text</code>.</li>
                       </ol>
                     </li>
-                    <li id="step2B.iii">Return the <code>accumulated text</code>.</li>
+                    <li id="step2B.iii">Return the <code>accumulated text</code> if it is not the empty string ("").</li>
                   </ol>
                 </li>
               </ul>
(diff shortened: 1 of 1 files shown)


MelSumner (approved):
This makes sense. Thank you for your PR. 👍

accdc (approved):
Looks good to me.

PR #122, opened 2021-04-12
suggested simplification

### LINKED ISSUE #96: Suggested simplification
So, I'm reading the Chrome code, and looking at 2c-2e and I'm thinking a big simplification would be to reorder it so that the current 2E comes before 2C, then 2D.

Right now, 2C is basically saying, maybe do 2E now if we have an aria-label, and then stop. But if if there is no aria-label, then it will let native markup attributes take precedence over the value. It's very confusing that 2C is saying to do 2E first. So then why do we have that order at all?

It's much simpler and works just as well for any use cases I know of to say:
New 2C) If from descendant recursion and labelling another widget, use value if available [ was old 2E ]
New 2D) Otherwise, use non-empty aria-label if available [ was old 2C ]
New 2E) Otherwise, use native markup if available [ was old 2D ]

MelSumner:
Fixes #96 

If merged, this PR re-orders the computation steps as suggested:

> New 2C) If from descendant recursion and labelling another widget, use value if available [ was old 2E ]
New 2D) Otherwise, use non-empty aria-label if available [ was old 2C ]
New 2E) Otherwise, use native markup if available [ was old 2D ]

Also, if this is merged, all issues marked with "2E" or other references will need to be updated.

***
<a href="https://raw.githack.com/w3c/accname/issue-96-computation-steps/index.html">Preview</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -294,7 +294,26 @@
                 </pre>
               </details></div>
             </li>
-            <li id="step2C">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> <a class="termref">attribute</a> whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
+            <li id="step2C">Otherwise, if the <code>current node</code> is a control embedded within the label (e.g. the <code>label</code> element in HTML or any element directly referenced by <code>aria-labelledby</code>) for another <a class="termref">widget</a>, where the user can adjust the embedded control's value, then include the embedded control as part of the text alternative in the following manner:
+              <ul>
+                <li>If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
+                <li>If the embedded control has role menu <a class="role-reference" href="#button">button</a>, return the text alternative of the button.</li>
+                <li>If the embedded control has role <a class="role-reference" href="#combobox">combobox</a> or <a class="role-reference" href="#listbox">listbox</a>, return the text alternative of the chosen <a class="role-reference" href="#option">option</a>.</li>
+                <li>If the embedded control has role <a class="role-reference" href="#range">range</a> (e.g., a <a class="role-reference" href="#spinbutton">spinbutton</a> or <a class="role-reference" href="#slider">slider</a>):
+                  <ul>
+                    <li>If the <code>aria-valuetext</code> property is present, return its value, </li>
+                    <li>Otherwise, if the <code>aria-valuenow</code> property is present, return its value,</li>
+                    <li>Otherwise, use the value as specified by a host language attribute. </li>
+                  </ul>
+                </li>
+              </ul>
+              <div><details>
+                <summary>Example:</summary>
+                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the user has entered "5" for the embedded <a class="role-reference" href="#textbox">textbox</a>, the complete label is "Flash the screen 5 times", e.g.: </p>
+                <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
+              </details></div>
+            </li>
+            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> <a class="termref">attribute</a> whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
               <ul>
                 <li>If traversal of the <code>current node</code> is due to recursion <strong>and</strong> the <code>current node</code> is an embedded control as defined in step 2E, ignore <code>aria-label</code> and skip to rule 2E.</li>
                 <li>Otherwise, return the value of <code>aria-label</code>.</li>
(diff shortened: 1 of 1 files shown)


MelSumner:
@michael-n-cooper so my local preview is showing the numbering happening properly, but the preview here in the PR is not showing the same thing (it's showing 2, then 1,2,3 etc instead of showing 2, then A,B,C etc). Is this a known thing or am I doing something incorrectly here?

jnurthen:
> @michael-n-cooper so my local preview is showing the numbering happening properly, but the preview here in the PR is not showing the same thing (it's showing 2, then 1,2,3 etc instead of showing 2, then A,B,C etc). Is this a known thing or am I doing something incorrectly here?

Known issue. PR-preview doesn't include our style sheets. 

If it causes a problem understanding the PR (and here it probably does) adding a link to a https://raw.githack.com/ version of the PR would be useful

aleventhal (approved):
Thank you.

We may want to do other edits, but I like having this pull request being only the rearrangement of these items.

MelSumner:
Also @jcsteh if you could look at this change too, thank you!

cookiecrook (approved):
From the diff, it wasn't immediately clear to me that this was a rearrangement-only edit. Since it's now clear that this moves the embedded control section in front of the labeling attr/el section (e.g. `title` attr or `label` element in HTML) this seems fine to me, too. Thanks Melanie.

index.html
```diff
@@ -294,7 +294,26 @@ <h4>Computation steps</h4>
...
+                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the user has entered "5" for the embedded <a class="role-reference" href="#textbox">textbox</a>, the complete label is "Flash the screen 5 times", e.g.: </p>
+                <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
+              </details></div>
+            </li>
+            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> <a class="termref">attribute</a> whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
               <ul>
                 <li>If traversal of the <code>current node</code> is due to recursion <strong>and</strong> the <code>current node</code> is an embedded control as defined in step 2E, ignore <code>aria-label</code> and skip to rule 2E.</li>
```
jcsteh, line 318: This new 2D refers to embedded controls in 2E, but embedded controls are now covered by 2C, making this text incorrect. Do we even need this piece of text given that 2C now precedes it?

index.html
```diff
@@ -294,7 +294,26 @@ <h4>Computation steps</h4>
                 </pre>
               </details></div>
             </li>
-            <li id="step2C">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> <a class="termref">attribute</a> whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
+            <li id="step2C">Otherwise, if the <code>current node</code> is a control embedded within the label (e.g. the <code>label</code> element in HTML or any element directly referenced by <code>aria-labelledby</code>) for another <a class="termref">widget</a>, where the user can adjust the embedded control's value, then include the embedded control as part of the text alternative in the following manner:
```
jcsteh, line 297: This may be something to address separately, but I worry that it's confusing that we say "include the embedded control as part of the text alternative" rather than "return" like we do everywhere else. Although we say "otherwise" in later steps, "include" here could be misconstrued to suggest that we should proceed with the following steps even if this step 2C applied.

jcsteh (commented):
Do we have examples (obscure/unrealistic as they may be) for which the actual behaviour would change here?

We explicitly say the embedded control stuff should take precedence over aria-label. That's now even clearer with the reordering. But this leads me to wonder: why should aria-labelledby (2B) be any different? If I read this correctly, aria-labelledby takes precedence over the value even for a label-embedded control, which seems wrong and neither Firefox nor Chrome do it:

`data:text/html,<label><input type="checkbox">Flash the screen <input type="text" value="5" aria-labelledby="blah"> times</label><div id="blah">Blah</div>`

While I don't want to overload a single PR with too many changes, since we're reordering here anyway, would it make sense to move the old 2E to 2B? So we'd have:

New 2B) If from descendant recursion and labelling another widget, use value if available [ was old 2E ]
New 2C) Otherwise, traverse aria-labelledby if available [ was old 2B ]
New 2D) Otherwise, use non-empty aria-label if available [ was old 2C ]
New 2E) Otherwise, use native markup if available [ was old 2D ]

MelSumner:
> Do we have examples (obscure/unrealistic as they may be) for which the actual behaviour would change here?
> 
> We explicitly say the embedded control stuff should take precedence over aria-label. That's now even clearer with the reordering. But this leads me to wonder: why should aria-labelledby (2B) be any different? If I read this correctly, aria-labelledby takes precedence over the value even for a label-embedded control, which seems wrong and neither Firefox nor Chrome do it:
> 
> `data:text/html,<label><input type="checkbox">Flash the screen <input type="text" value="5" aria-labelledby="blah"> times</label><div id="blah">Blah</div>`
> 
> While I don't want to overload a single PR with too many changes, since we're reordering here anyway, would it make sense to move the old 2E to 2B? So we'd have:
> 
> New 2B) If from descendant recursion and labelling another widget, use value if available [ was old 2E ]
> New 2C) Otherwise, traverse aria-labelledby if available [ was old 2B ]
> New 2D) Otherwise, use non-empty aria-label if available [ was old 2C ]
> New 2E) Otherwise, use native markup if available [ was old 2D ]

Tracking these proposed edits in https://github.com/w3c/accname/issues/125

index.html
MelSumner: Tracked with issue https://github.com/w3c/accname/issues/126

PR #137, opened 2021-08-06
Updated spec text to reflect the processing of hidden elements when c…

### LINKED ISSUE #57: When is hidden content taken into calculation of name and description?
Unfortunately for me (as well as for the browser manufacturers and developers of assistive technology: https://github.com/FreedomScientific/VFO-standards-support/issues/251) the procedure for determining labeling and description is too complex (https://w3c.github.io/accname/#terminology) and I would therefore like to know how to proceed with hidden content according to the specification. 

For which label types is hidden content output valid?
* aria-labelledby 
* aria-describedby
* label for
* enclosing label
* legend in fieldset
* figcaption in figure
* caption in table
* determining accessible name and description for elements within a Live Region
* ...

When does it apply?
* Label or description itself is hidden
* Descendant elements (of label or description) are hidden
* Ancestor elements (of label or description) are hidden

If I'm not the only one who doesn't see this directly from the specification, I'd ask you to be more precise in the specification and possibly give some examples of when hidden content is output and when not.

See also: https://github.com/w3c/aria-practices/issues/1136

accdc:
…omputing description to match intent as discussed.

Closes #57

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/137.html" title="Last updated on Sep 23, 2021, 5:44 PM UTC (1e6fd4c)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/137/bbf8eb2...1e6fd4c.html" title="Last updated on Sep 23, 2021, 5:44 PM UTC (1e6fd4c)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -265,7 +265,7 @@
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">if computing a <a class="termref" data-lt="accessible name">name</a>, or if computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is not <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2. If computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is <a class="termref">hidden</a>, however, compute the text alternative of the <code>current node</code> by returning its inner text content. Set the <code>result</code> to that text alternative.</li>
                         <li id="step2B.ii.c">Append the <code>result</code>, with a space, to the <code>accumulated text</code>.</li>
                       </ol>
                     </li>
--- a/index.html
+++ b/index.html
@@ -269,7 +269,7 @@
                             Compute the text alternative of the <code>current node</code> using the following steps: 
                             <ul>
                                 <li>If the <code>current node</code> is <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>, recursively skip to step 2.</li>
-                                <li>If the <code>current node</code> is <a class="termref">hidden</a>:
+                                <li>If the <code>current node</code> is <strong?not</strong> <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>:
                                   <ol>
                                     <li>If the <code>current node</code> has an <a class="pref">aria-label</a> attribute whose value, when trimmed of white space, is not the empty string, return the value of <a class="pref">aria-label</a>.</li>
                                     <li>Otherwise, return the {{HTMLElement/innerText}} of the node.</li>
--- a/index.html
+++ b/index.html
@@ -268,7 +268,7 @@
                         <li id="step2B.ii.b">
                             Compute the text alternative of the <code>current node</code> using the following steps: 
                             <ul>
-                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, recursively skip to step 2.</li>
+                                <li>If the <code>current node</code> is <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>, recursively skip to step 2.</li>
                                 <li>If the <code>current node</code> is <a class="termref">hidden</a>:
                                   <ol>
                                     <li>If the <code>current node</code> has an <a class="pref">aria-label</a> attribute whose value, when trimmed of white space, is not the empty string, return the value of <a class="pref">aria-label</a>.</li>
--- a/index.html
+++ b/index.html
@@ -269,7 +269,7 @@
                             Compute the text alternative of the <code>current node</code> using the following steps: 
                             <ul>
                                 <li>If the <code>current node</code> is <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>, recursively skip to step 2.</li>
-                                <li>If the <code>current node</code> is <strong?not</strong> <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>:
+                                <li>If the <code>current node</code> is <strong>not</strong> <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>:
                                   <ol>
                                     <li>If the <code>current node</code> has an <a class="pref">aria-label</a> attribute whose value, when trimmed of white space, is not the empty string, return the value of <a class="pref">aria-label</a>.</li>
                                     <li>Otherwise, return the {{HTMLElement/innerText}} of the node.</li>
--- a/index.html
+++ b/index.html
@@ -265,7 +265,13 @@
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">if computing a <a class="termref" data-lt="accessible name">name</a>, or if computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is not <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2. If computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is <a class="termref">hidden</a>, however, compute the text alternative of the <code>current node</code> by returning its inner text content. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">
+                            <ul>
+                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2.</li>
+                                <li>If the <code>current node</code> is <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> by returning {{HTMLElement/innerText}} of the node.</li>
+                            </ul>
+                            Set the <code>result</code> to the computed text alternative from this step.
+                        </li>
                         <li id="step2B.ii.c">Append the <code>result</code>, with a space, to the <code>accumulated text</code>.</li>
                       </ol>
                     </li>
--- a/index.html
+++ b/index.html
@@ -269,7 +269,12 @@
                             Compute the text alternative of the <code>current node</code> using the following steps: 
                             <ul>
                                 <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, recursively skip to step 2.</li>
-                                <li>If the <code>current node</code> is <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> by returning {{HTMLElement/innerText}} of the node.</li>
+                                <li>If the <code>current node</code> is <a class="termref">hidden</a>:
+                                  <ol>
+                                    <li>If the <code>current node</code> has an <a class="pref">aria-label</a> attribute whose value, when trimmed of white space, is not the empty string, return the value of <a class="pref">aria-label</a>.</li>
+                                    <li>Otherwise, return the {{HTMLElement/innerText}} of the node.</li>
+                                  </ol>
+                                </li>
                             </ul>
                             Set the <code>result</code> to the computed text alternative from this step.
                         </li>
--- a/index.html
+++ b/index.html
@@ -266,8 +266,9 @@
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
                         <li id="step2B.ii.b">
+                            Compute the text alternative of the <code>current node</code> using the following steps: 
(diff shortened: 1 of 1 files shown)


cookiecrook:
I find this line very difficult to parse:

> If computing a name, or if computing a description and the current node is not hidden, compute the text alternative of the current node beginning with step 2. If computing a description and the current node is hidden, however, compute the text alternative of the current node by returning its inner text content. Set the result to that text alternative.

I tried rewriting it for clarity:

> Compute the text alternative using the following steps:
> - When computing a name, compute the text alternative of the current node beginning with step 2 (this should point to a named step rather than a numbered step). 
> - When computing a description:
>   - If the current node is not hidden, compute the text alternative of the current node beginning with step 2 (ditto re: numbered step). 
>   - If the current node is hidden, compute the text alternative of the current node by returning the value of `element.innertext`.
> 
> Set the result to that text alternative.

The last line isn't changed as part of this PR, but it could benefit from further clarification too.

jnurthen:
Referencing {{HTMLElement/innerText}} will reference innerText in the HTML spec

cookiecrook (commented):
@accdc Assuming you accept my editorial change proposal of your prose, and @aleventhal's suggestion to simplify (keeping this step the same for both name and description), then I will float this proposal further with other WebKit engineers. The accessibility engineers have already reviewed, but there are more stakeholders wrt `innerText.`

Notes: 
- Any remaining differences between the browser implementations of `innerText` should probably be addressed as separate issues. That's part of the value of this change. It replies on the HTML definition of intterText, rather than defining it in the ARIA spec.
- I have not addressed the numbering issues (prose "step 2" and ID "step2B.ii.b") because those will be addressed in #139

index.html
```diff
@@ -265,7 +265,7 @@ <h4>Computation steps</h4>
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">if computing a <a class="termref" data-lt="accessible name">name</a>, or if computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is not <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2. If computing a <a class="termref" data-lt="accessible description">description</a> and the <code>current node</code> is <a class="termref">hidden</a>, however, compute the text alternative of the <code>current node</code> by returning its inner text content. Set the <code>result</code> to that text alternative.</li>
```
cookiecrook, line 268:
```suggestion
                        <li id="step2B.ii.b">
                            <ul>
                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2.</li>
                                <li>If the <code>current node</code> is <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> by returning {{HTMLElement/innerText}} of the node.</li>
                            </ul>
                            Set the <code>result</code> to the computed text alternative from this step.
                        </li>
```

cookiecrook (commented):
Looks like the change proposal didn't come through the first time.

index.html
```diff
@@ -265,7 +265,13 @@ <h4>Computation steps</h4>
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">
+                            <ul>
+                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2.</li>
+                                <li>If the <code>current node</code> is <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> by returning {{HTMLElement/innerText}} of the node.</li>
```
cookiecrook, line 271:
```suggestion
                                <li>If the <code>current node</code> is <a class="termref">hidden</a>:
                                  <ol>
                                    <li>If the <code>current node</code> has an <a class="pref">aria-label</a> attribute whose value, when trimmed of white space, is not the empty string, return the value of <a class="pref">aria-label</a>.</li>
                                    <li>Otherwise, compute the text alternative of the <code>current node</code> by returning the {{HTMLElement/innerText}} of the node.</li>
                                  </ol>
                                </li>
```

cookiecrook: Pending PR change to address @jaragunde's comment in #144

index.html
```diff
@@ -265,7 +265,13 @@ <h4>Computation steps</h4>
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">
+                            <ul>
+                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, compute the text alternative of the <code>current node</code> beginning with step 2.</li>
```
cookiecrook, line 270:
```suggestion
                            Compute the text alternative of the <code>current node</code> using the following steps: 
                            <ul>
                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, recursively skip to step 2.</li>
```

cookiecrook (approved):

cookiecrook:
FWIW, I removed my "approved" review, since I rewrote a significant portion of this change, but if we get two other reviewers, I think it looks okay.

aleventhal (commented):
Looks like there might be two places that we need to mention aria-label? There's another place that is adding the innerText language, but it doesn't mention aria-label yet.

aleventhal:
Looks good with @cookiecrook 's change.

cookiecrook:
> Looks like there might be two places that we need to mention aria-label? There's another place that is adding the innerText language, but it doesn't mention aria-label yet.

I think it's just on Line 271. Where else are you seeing it?

index.html
cookiecrook: m. editorial

aleventhal (approved):

jnurthen:
@jcsteh - can you please take a look at this

jcsteh:
Name/description calculation hurts my head. 🤯

Overall, I agree this simplifies things and makes sense for that reason. One thing that occurs to me: while the differences in output when *descendants* use different methods of hiding have been discussed, there will also be significant differences when the target node itself uses different methods of hiding. For example:

```
<div id="label" hidden>
  visible <span hidden>hidden</span>
</div>
<button aria-labelledby="label"></button>
```
Button label will be "visible hidden"

```
<div id="label" style="visibility: hidden;">
  visible <span hidden>hidden</span>
</div>
<button aria-labelledby="label"></button>
```
Button label will be ""

```
<div id="label" aria-hidden="true">
  visible <span hidden>hidden</span>
</div>
<button aria-labelledby="label"></button>
```
Button label will be "visible"

Currently, Firefox returns the same label for all of these: "visible hidden"

That said, I guess this is pretty similar to the "maybe not what authors expect" situation for descendants and consensus seems to be that this is an acceptable compromise. I just thought it worth raising in case it hadn't been considered yet (and is worthy of separate consideration).

aleventhal:
We found something that might be a problem. In this example, only the display:none has an innerText, but I would have expected visibility:hidden innerText to remain intact.

```
<div id="dn" style="display:none">disp-none</div>
<div id="vh" style="visibility:hidden">visi-hidden</div>
```

Therefore, this PR will break labels where authors thought that visibility:hidden was an acceptable way to hide a label.

aleventhal (changes requested):
Removing my approval until we discuss the fact that visibilitity:hidden on the target of an aria-labelledby will lead to an empty name.

accdc:
Thanks for the heads up. I'm not sure if it is possible to resurrect, we will need to talk it over at the next meeting. I may not be available for the one on the 16th due to a conflict though.

index.html
```diff
@@ -265,7 +265,19 @@ <h4>Computation steps</h4>
                     <li id="step2B.ii">For each IDREF:
                       <ol>
                         <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
-                        <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
+                        <li id="step2B.ii.b">
+                            Compute the text alternative of the <code>current node</code> using the following steps: 
+                            <ul>
+                                <li>If the <code>current node</code> is <strong>not</strong> <a class="termref">hidden</a>, recursively skip to step 2.</li>
```
jnurthen, line 271:
```suggestion
                                <li>If the <code>current node</code> is <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>, recursively skip to step 2.</li>
```

index.html
```diff
@@ -265,7 +265,19 @@ <h4>Computation steps</h4>
...
+                                <li>If the <code>current node</code> is <a class="termref">hidden</a>:
```
jnurthen, line 272:
```suggestion
                                <li>If the <code>current node</code> is <strong?not</strong> <a href="https://html.spec.whatwg.org/multipage/rendering.html#being-rendered">being rendered</a>:
```

jnurthen (changes requested):
Would these changes fix the visibility:hidden issue - it uses the same text from HTML to determine if something is rendered so would only use innerText for those usages.

index.html
aleventhal: @jnurthen so we would treat visibility:hidden as before, and display:none would get the new innerText treatment?

If that's what we decide to do, then I think we might want to call that out explicitly. It feels like the HTML definition would require interpretation by people who understand how browsers work.

jnurthen: yeah - although having suggested this I since realised that this doesn't account for aria-hidden which is rendered but should probably fit into the non rendered category that uses innerText.

cyns: @cookiecrook does this solve your implementation problem? Is HTML's definition of "rendered" what you need? 

@aleventhal are you good with this if we add examples for visibility:hidden and display:none? Maybe also html hidden attribute, opacity:0 and offscreen. I think it's easier to understand this from examples than from the algorithm, plus they can be the start of test files.

aleventhal: That sounds good in general. If we're going to be complete, there is also content-visibility:hidden, I suppose.
I'm assuming that everyone agrees opacity:0 and offscreen are treated as not hidden.

scottaohara: yes.  opacity: 0 and off-screen positioning (via position, negative text indent, etc) should not be treated the same as hidden.

jcsteh (commented):
I assume this change means that we still need to handle ARIA attributes, etc. in visibility: hidden targets of aria-labell/describedby? For example:

```
<div id="l3" hidden>l3</div>
<div id="label" style="visibility: hidden;">
  l1
  <span aria-hidden="hidden">l2</span>
  <span aria-labelledby="l3"></span>
</div>
<button aria-labelledby="label"></button>
```

If I read this right, the expected label for the button is "l1 l3". in Firefox, that's tricky because visibility: hidden subtrees get removed from the a11y tree. That's not a new bug - we need to fix it to fully comply with the current spec too - but I wanted to clarify the intent here.

On the other hand, if you do this:

```
<div id="l3" hidden>l3</div>
<div id="label">
  <span style="visibility: hidden;">
    l1
    <span aria-hidden="hidden">l2</span>
    <span aria-labelledby="l3"></span>
  </span>
</div>
<button aria-labelledby="label"></button>
```

If I read this right, the expected label for the button will be empty, since we don't differentiate "being rendered" from "hidden" except for direct targets of aria-labell/describedby.

cookiecrook:
@jcsteh I'm not sure if you are expressing a block/concern in there, or if you are just noting potential implementation complexity.

cookiecrook:
@jcsteh I believe your understanding is correct. The follow up change was proposed by @aleventhal to account for an authoring practice he anticipates may cause breakage if we used innerText without exception.

I think both Aaron’s and Jamie’s points have merit, but they are at odds with each other. Maybe a compromise can be reached.

Perhaps this topic is worthy of more follow-up discussion at an upcoming ARIA meeting. Ideally, @accdc, @joanmarie, @jcsteh, and @aleventhal should attend. I’d like to be there too.

aleventhal:
Is this proposal dead now?
At least for Chrome, we no longer need it. We found a way to work around our acc name implementation's complexity and follow the current spec without using innerText.

accdc:
[test.zip](https://github.com/w3c/accname/files/7479361/test.zip)

accdc:
As we spoke of earlier, I've attached an example that would need to be solvable by whatever we decide for this issue going forward.

cookiecrook:
@aleventhal wrote:

> Is this proposal dead now?

No. From the ARIA call (last week IIRC?), we left this with Aaron to make a change suggestion on the PR that either didn't include innerText, or had some more exceptions to the innerText option.

jnurthen:
@aleventhal should this be closed now?

jnurthen:
Discussed in Feb 3 2022 call. Closing as overcome

PR #140, opened 2021-08-16
Clarify that descendant names should be joined without a space in step 2F.iii.c

### LINKED ISSUE #15: Step 2.F.iii.c doesn't say if result should be appended with or without a space
Step 2.F.iii.c doesn't say if result should be appended with or without a space. The rest of the algorithm is more specific.

Related to this are the tests like:
```
  <div id="test" role="link" tabindex="0">
   ...
    (<span>Q</span><span>E</span><span>D</span>)
   ...
  </div>
```
Some implementations are turning that portion into "( QED )" and others into "( Q E D )". The expectation is "(QED)" (sans spaces). That seems to imply that 2.F.iii.c should be an append without spaces. If that is indeed the case, what would be the expectation if the above snippet instead said:

```
    (<span aria-label="Quod">Q</span><span aria-label="erat">E</span><span aria=label="demonstrandum">D</span>)
```

What if buttons were used instead of spans? What if divs were used instead of spans? What if there's a mix?

Edit: What if the middle span were instead a text input. We have this test case:

```
if given
  <input type="radio" id="test">
  <label for="test">foo<input type="text" value="bar">baz</label>
then the accessible name of the element with id of "test" is "foo bar baz"
```
I'm pretty sure the part of the algorithm which dictates calculating name from content of the label is also 2.F. And in that case our expectation is the insertion of a space.

I realize the whitespace situation is a mess, but clarification on the above would be helpful in terms of fixing implementations -- assuming they indeed need to be fixed (i.e. because our "(QED)" expectations are correct).

kasperisager:
Resolves #15.

***
<a href="https://pr-preview.s3.amazonaws.com/kasperisager/accname/pull/140.html" title="Last updated on Aug 16, 2021, 10:36 AM UTC (9d45e83)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/140/bbf8eb2...kasperisager:9d45e83.html" title="Last updated on Aug 16, 2021, 10:36 AM UTC (9d45e83)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -345,7 +345,7 @@
                   <ol>
                     <li id="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
                     <li id="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with step 2.  Set the <code>result</code> to that text alternative.</li>
-                    <li id="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="step2F.iii.c">Append the <code>result</code>, without a space, to the <code>accumulated text</code>. </li>
                   </ol>
                 </li>
                 <li id="step2F.iv">Return the <code>accumulated text</code>.</li>


MelSumner (changes requested):
Based on https://github.com/w3c/accname/issues/15#issuecomment-1004524133 - I wonder if it should be 

"Append the result to the accumulated text, using the presence or absence of whitespace between the inline elements as the deciding factor for whether or not to include a space." 

What do you think?

accdc:
I agree, it would be important to take that into account. I'm having trouble viewing the proposed spec text change here, can someone please post it in plain text for review?

kasperisager:
That's definitely important to account for, yes! My only concern would be ensuring an unambiguous definition of such whitespace.

As I'm no longer involved in any W3C work, I'll ping @Jym77 as this might be something he'd be interested in picking up. Hi Jean-Yves! 👋😄

Issue #148, opened 2022-02-08
Clarification of behavior for "native host language text alternative element" in step 2A


jaragunde:
This is the relevant piece of text I would like to discuss, in <a href="https://w3c.github.io/accname/#step2A">step 2A</a>:

> If the current node is [hidden](https://www.w3.org/TR/wai-aria-1.2/#dfn-hidden) and is not directly referenced by aria-labelledby or aria-describedby, nor directly referenced by a native host language text alternative [element](https://www.w3.org/TR/wai-aria-1.2/#dfn-element) (e.g. label in HTML) or [attribute](https://www.w3.org/TR/wai-aria-1.2/#dfn-attribute), return the empty string.

While we have discussed a lot about the meaning on the clause referencing aria-labelledby/describedby in #57, I don't really understand what the other part means, about "directly referenced by a native host language text alternative element". I can think of this example:

```html
<label for='box' hidden>foo</label>
<input id='box'>
```
In the example above, all main user agents produce an empty accessible name, so they wouldn't be following the spec text, or the text might not refer to this case.

A more convoluted example:
```html
<label aria-labelledby="someHiddenElementElsewhere"><input></label>
<div id="someHiddenElementElsewhere" hidden>foo!</div>
```
This produces the accname "foo!" for the input in Chrome and WebKit, and empty in Firefox. But I'm not sure that the "directly referenced by a native host language text alternative element" text is meant to address this case, I see it more as a combination of steps [2B](https://w3c.github.io/accname/#step2B) and [2E](https://w3c.github.io/accname/#step2E) working together.

I would appreciate any thoughts on this. Thanks!

scottaohara:
Interesting that with the second example the input gets its name.  Not necessarily opposed to that, but seems an error correction since `<label>` would generally be an element that is prohibited by author naming via aria attributes.

i do see value in having a hidden `<label>` to name a form control.  but i don't think that's 'expected' to work right now?

jaragunde:
> i do see value in having a hidden <label> to name a form control. but i don't think that's 'expected' to work right now?

Not expected to work (certainly not implemented), but UAs may already be hiding the label next to the control by default, to prevent double-speech of the label + the control. We certainly do that in Chromium in the most basic cases (checkboxes and text controls, IIRC). Screen readers do their fair share of correction, too.

That was an aside, anyway... In your opinion, that piece of text refers exactly to this, a hidden label being used to name a control?

PR #150, opened 2022-03-15
Specify behavior when following a hidden subtree via aria-labelledby …

### LINKED ISSUE #57: When is hidden content taken into calculation of name and description?

jaragunde:
…or -describedby.

This proposal tries to bring the spec closer to the actual behavior of the different user agents. In general, user agents expose hidden content when following an aria-labelledby or -describedby relation into a hidden subtree. There are differences between engines, though.

This is a summary of what user agents will need to change to comply with this text update:

* Firefox in generally compliant with the proposed wording, but it might need checking for special cases.
* Chrome would need to change the behavior for aria-hidden. Example:
  ```html
  <style>.hidden {display:none;}</style>
  <input id="el1" role="button" aria-labelledby="el2" />
  <span id="el2" class="hidden">
    <span id="el3" aria-hidden="true">hello</span>
  </span>
  ```
  Name for the input should be "hello", currently empty in Chrome.
* WebKit would need to unify behavior between different types of hidden, because it currently produces different names when using `display:none` and `visibility:hidden`. Example:
  ```html
  <style>.hidden {visibility:hidden;}</style>
  <input id="el1" role="button" aria-labelledby="el2" />
  <span id="el2" class="hidden">
    <span id="el3" class="hidden">hello</span>
  </span>
  ```
  Name for the input should be "hello", currently empty in WebKit. It would be "hello" if `display:none` was used instead.

Thanks in advance for your time and feedback.

Closes #57.

***
<a href="https://pr-preview.s3.amazonaws.com/jaragunde/accname/pull/150.html" title="Last updated on Apr 20, 2022, 2:14 PM UTC (128349a)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/150/b3b5558...jaragunde:128349a.html" title="Last updated on Apr 20, 2022, 2:14 PM UTC (128349a)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -278,7 +278,7 @@
   &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
 &lt;/element2&gt;</code>
                 </pre>
-                <p>Alternatively, <code>element1</code> has no <a class="termref">accessible name</a> if <code>element3</code> is hidden and it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, but <code>element2</code> is not hidden.</p>
+                <p>Conversely, <code>element1</code> has no <a class="termref">accessible name</a> if <code>element3</code> is hidden and it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, but <code>element2</code> is not hidden.</p>
                 <pre class="example highlight"><code>&lt;element1 id="el1" role="button" aria-labelledby="el2" /&gt;
 &lt;element2 id="el2"&gt;
   &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
--- a/index.html
+++ b/index.html
@@ -249,10 +249,41 @@
         <li id="step1">Set the <code>root node</code> to the given element, the <code>current node</code> to the <code>root node</code>, and the <code>total accumulated text</code> to the empty string (""). If the <code>root node</code>'s role prohibits naming, return the empty string ("").</li>
         <li id="step2">Compute the text alternative for the <code>current node</code>:
           <ol>
-            <li id="step2A">If the <code>current node</code> is <a class="termref">hidden</a> <strong>and is not</strong> directly referenced by <code>aria-labelledby</code> or <code>aria-describedby</code>, nor directly referenced by a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML) or <a class="termref">attribute</a>, return the empty string.
+            <li id="step2A">If the <code>current node</code> is <a class="termref">hidden</a> <strong>and is</strong>:
+            <ul>
+              <li><strong>Not</strong> part of an <code>aria-labelledby</code> or <code>aria-describedby</code> traversal, where the node directy referenced by that relation was hidden.</li>
+              <li><strong>Nor</strong> part of a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML) or <a class="termref">attribute</a> traversal, where the root of that traversal was hidden.</li>
+            </ul>
+            Return the empty string.
+              <div><details>
+                <summary>Comment:</summary>
+                <p>It's important to clarify the broad definition of hidden for the purposes of accessible name calculation:
+                <ul>
+                  <li>Nodes with CSS properties <code>display:none</code>, <code>visibility:hidden</code>, <code>visibility:collapse</code> or <code>content-visibility:hidden</code>: They are considered hidden, as they match the guidelines "not perceivable" and "explicitly hidden".</li>
+                  <li>Nodes with CSS properties <code>opacity:0</code> or <code>filter:opacity(0%)</code>, or similar SVG mechanisms: They are not considered hidden. Text hidden with these methods can still be selected or copied, and user agents still expose it in their accessibility trees.</li>
+                  <li>Nodes with the <code>aria-hidden="true"</code> property: it is considered hidden, matching the "explicitly hidden" guideline.</li>
+                  <li>Nodes hidden off screen or behind another object: they are not considered hidden. They are exposed in the accessibility tree and they can even name on-screen objects.</li>
+                </ul></p>
+              </details></div>
               <div><details>
                 <summary>Comment:</summary>
                 <p>By default, <a class="termref">assistive technologies</a> do not relay hidden information, but an author can explicitly override that and include hidden text as part of the <a class="termref">accessible name</a> or <a class="termref">accessible description</a> by using <code>aria-labelledby</code> or <code>aria-describedby</code>. </p>
+              </details></div>
+              <div><details>
+                <summary>Example:</summary>
+                <p>The following examples show the meaning of the clause "Not part of an <code>aria-labelledby</code> or <code>aria-describedby</code> traversal, where the node directy referenced by that relation was hidden.".</p>
+                <p>First, <code>element1</code>'s <a class="termref">accessible name</a> is "hello" because, although <code>element3</code> is hidden, it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, which is hidden too.</p>
+                <pre class="example highlight"><code>&lt;element1 id="el1" role="button" aria-labelledby="el2" /&gt;
+&lt;element2 id="el2" class="hidden"&gt;
+  &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
+&lt;/element2&gt;</code>
+                </pre>
+                <p>Alternatively, <code>element1</code> has no <a class="termref">accessible name</a> if <code>element3</code> is hidden and it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, but <code>element2</code> is not hidden.</p>
+                <pre class="example highlight"><code>&lt;element1 id="el1" role="button" aria-labelledby="el2" /&gt;
+&lt;element2 id="el2"&gt;
+  &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
+&lt;/element2&gt;</code>
+                </pre>
               </details></div>
             </li>
             <li id="step2B">Otherwise:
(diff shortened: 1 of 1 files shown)


jaragunde:
I've setup my account as requested, feel free to "revalidate". Thanks!

accdc (approved):
Personally I like it and approve. Pending review by @MelSumner and @cookiecrook of course. Hopefully we can discuss this at the next meeting to finally get this pulled in.

MelSumner:
I'm looking at these use cases in a [CodePen](https://codepen.io/melsumner/pen/RwxKjjp/0e79255d0517a899f2c4a5b29e2d0e9a?editors=1000), apologies for the review delay.

index.html
```diff
@@ -249,11 +249,42 @@ <h4>Computation steps</h4>
...
+              <div><details>
+                <summary>Example:</summary>
+                <p>The following examples show the meaning of the clause "Not part of an <code>aria-labelledby</code> or <code>aria-describedby</code> traversal, where the node directy referenced by that relation was hidden.".</p>
+                <p>First, <code>element1</code>'s <a class="termref">accessible name</a> is "hello" because, although <code>element3</code> is hidden, it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, which is hidden too.</p>
+                <pre class="example highlight"><code>&lt;element1 id="el1" role="button" aria-labelledby="el2" /&gt;
+&lt;element2 id="el2" class="hidden"&gt;
+  &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
+&lt;/element2&gt;</code>
+                </pre>
+                <p>Alternatively, <code>element1</code> has no <a class="termref">accessible name</a> if <code>element3</code> is hidden and it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, but <code>element2</code> is not hidden.</p>
```
jcsteh, line 281: nit: Perhaps "Conversely" or "In contrast" instead of "Alternatively". "Alternatively" is a totally acceptable word here, but it's possible to misinterpret it as "this alternative example demonstrates the same behaviour", when it's actually the reverse behaviour. The rest of the sentence makes that clear, but I thought it worth suggesting anyway.

jcsteh (approved):
This looks good to me. I always find this spec hurts my brain a lot and it takes me forever to figure out what behaviour exactly is changing. 😳 For the sake of absolute clarity, I assume the first example you've added demonstrates the difference between the previous and proposed new behaviours? With the previous behaviour (as specified), the label for element1 would be empty because element3 is hidden and not directly referenced.

jaragunde:
@jcsteh: Exactly, that's the difference between the previous wording and the new one.

index.html
jaragunde: Done, thanks for the suggestion!

jaragunde:
@MelSumner: thanks for the CodePen, it's very useful!

MelSumner (approved):
seems like we are agreed on these changes, thank you for the submission.

MelSumner:
Not sure why it was marking some changes as conflicted, but I think they are resolved now.

jnurthen:
@MelSumner I thought we were waiting for @cookiecrook and @cyns reviews

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
         <li id="step1">Set the <code>root node</code> to the given element, the <code>current node</code> to the <code>root node</code>, and the <code>total accumulated text</code> to the empty string (""). If the <code>root node</code>'s role prohibits naming, return the empty string ("").</li>
         <li id="step2">Compute the text alternative for the <code>current node</code>:
           <ol>
-            <li id="step2A">If the <code>current node</code> is <a class="termref">hidden</a> <strong>and is not</strong> directly referenced by <code>aria-labelledby</code> or <code>aria-describedby</code>, nor directly referenced by a native host language text alternative [=element=] (e.g. <code>label</code> in HTML) or [=attribute=], return the empty string.
+            <li id="step2A">If the <code>current node</code> is <a class="termref">hidden</a> <strong>and is</strong>:
+            <ul>
+              <li><strong>Not</strong> part of an <code>aria-labelledby</code> or <code>aria-describedby</code> traversal, where the node directy referenced by that relation was hidden.</li>
```
cookiecrook, line 274: Last two words ("_was_ hidden") seem to imply either transiency (e.g. element is no longer hidden) or that the element was the direct target of the hiding (e.g. this specific element is the actually hidden node rather than being hidden as a side effect of a hidden ancestor).

Perhaps "_is_ hidden" would be better here, since I understand the intent to be "currently hidden" and we don't care why/how it was hidden? If I'm misunderstanding, more clarification of the intent would be helpful.

I can't add a diff suggestion. Perhaps because it's already merged?

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
...
+              <li><strong>Nor</strong> part of a native host language text alternative <a class="termref">element</a> (e.g. <code>label</code> in HTML) or <a class="termref">attribute</a> traversal, where the root of that traversal was hidden.</li>
```
cookiecrook, line 275: ditto

```diff
- was hidden
+ is hidden
```

cookiecrook: "where the root of that traversal" may also be ambiguous. 

In this case, is "root" referring to the element labeled by this host language text alternative (e.g. an `<input>` labeled by `<label>` that is somewhere inside the overall computation)? Or is this referring to the root of the entire name computation traversal? I think the former because there may not be a time when a hidden node can be the root of the overall computation.

I labored over a diff for a few minutes but this may be similarly confusing. Sharing anyway as a straw man example. 
```diff
- where the root of that traversal
+ where the element referenced by the host language text alternative
```

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
...
+            </ul>
+            Return the empty string.
+              <div><details>
+                <summary>Comment:</summary>
+                <p>It's important to clarify the broad definition of hidden for the purposes of accessible name calculation:
+                <ul>
+                  <li>Nodes with CSS properties <code>display:none</code>, <code>visibility:hidden</code>, <code>visibility:collapse</code> or <code>content-visibility:hidden</code>: They are considered hidden, as they match the guidelines "not perceivable" and "explicitly hidden".</li>
+                  <li>Nodes with CSS properties <code>opacity:0</code> or <code>filter:opacity(0%)</code>, or similar SVG mechanisms: They are not considered hidden. Text hidden with these methods can still be selected or copied, and user agents still expose it in their accessibility trees.</li>
+                  <li>Nodes with the <code>aria-hidden="true"</code> property: it is considered hidden, matching the "explicitly hidden" guideline.</li>
```
cookiecrook, line 284: Editorial nits to match the other lines:

```diff
- it is
+ They are
```

There is also a mix of casings in They/they. I don't have a case preference other than consistency.

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
...
+                  <li>Nodes hidden off screen or behind another object: they are not considered hidden. They are exposed in the accessibility tree and they can even name on-screen objects.</li>
```
cookiecrook, line 285:
```diff
- Nodes hidden off screen
+ Nodes positioned off screen
```

cookiecrook:
```diff
- and they can even name on-screen objects.
+ and can label visible elements.
```

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
...
+                </ul></p>
+              </details></div>
               <div><details>
                 <summary>Comment:</summary>
                 <p>By default, <a class="termref">assistive technologies</a> do not relay hidden information, but an author can explicitly override that and include hidden text as part of the <a class="termref">accessible name</a> or <a class="termref">accessible description</a> by using <code>aria-labelledby</code> or <code>aria-describedby</code>. </p>
               </details></div>
+              <div><details>
+                <summary>Example:</summary>
+                <p>The following examples show the meaning of the clause "Not part of an <code>aria-labelledby</code> or <code>aria-describedby</code> traversal, where the node directy referenced by that relation was hidden.".</p>
```
cookiecrook, line 294:
```diff
- is hidden
+ was hidden
```

index.html
```diff
@@ -269,11 +269,42 @@ <h4>Computation steps</h4>
...
+                <p>First, <code>element1</code>'s <a class="termref">accessible name</a> is "hello" because, although <code>element3</code> is hidden, it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, which is hidden too.</p>
+                <pre class="example highlight"><code>&lt;element1 id="el1" role="button" aria-labelledby="el2" /&gt;
+&lt;element2 id="el2" class="hidden"&gt;
+  &lt;element3 id="el3" class="hidden"&gt;hello&lt;/element3&gt;
+&lt;/element2&gt;</code>
+                </pre>
+                <p>Conversely, <code>element1</code> has no <a class="termref">accessible name</a> if <code>element3</code> is hidden and it is part of an <code>aria-labelledby</code> traversal started in <code>element2</code>, but <code>element2</code> is not hidden.</p>
```
cookiecrook, line 301:
```diff
- but
+ because
```

cookiecrook (commented):
Mostly editorial with a few clarity questions. Overall I think it makes sense.

Since this is already merged, I could file a separate Issue/PR after we settle on a replacement or clarification for "root of the traversal"…

cookiecrook:
> @MelSumner I thought we were waiting for @cookiecrook and @cyns reviews

Merging provides additional motivation. 🤣

index.html
jaragunde: > Perhaps "is hidden" would be better here, since I understand the intent to be "currently hidden" and we don't care why/how it was hidden? 

Exactly this, I was not trying to express anything different. I should have used the present tense as in the original.

index.html
jaragunde: > is "root" referring to the element labeled by this host language text alternative ...?

Yes, that was my intention with that wording. I find the proposed diff fitting, it reuses the terminology used by the original wording which is more appropriate, although it's confusing in the same sense as the original was.

There is already an open issue about this: https://github.com/w3c/accname/issues/148

jaragunde:
Thanks for the review, I definitely agree with all your suggestions, some are even mistakes on my side caused by writing over and over.

To be more clear, I cannot think of a better suggestion for "root of the traversal". I would agree to revert it back to the term originally used, and keep track of what it means and how to enhance it in #148.

PR #159, opened 2022-05-16
experiment: merge main into stable 

MelSumner:
this PR is an experiment to merge main into stable (vs cherry-picking each commit). 

opened a PR because I am not 100% certain that this is okay from a process standpoint. 

Here's what I did to create this PR: 

- First, I created a branch of `stable`
- then I merged `main` into my branch
- then, I opened this PR to merge my branch into `stable`

But really, if this approach is acceptable, I probably could merge `main` directly into `stable` and push that. 

@jnurthen is this okay to do? What do you think of this approach?

***

### :boom: Error: 504 Gateway Timeout :boom: ###

[PR Preview](https://github.com/tobie/pr-preview#pr-preview) failed to build. _(Last tried on May 16, 2022, 2:25 PM UTC)_.

<details>
<summary>More</summary>

PR Preview relies on a number of web services to run. There seems to be an issue with the following one:

:rotating_light: [Spec Generator](https://www.w3.org/2015/labs/) - Spec Generator is the web service used to build specs that rely on ReSpec.

:link: [Related URL](https://labs.w3.org/spec-generator/?type=respec&url=https%3A%2F%2Fraw.githubusercontent.com%2Fw3c%2Faccname%2Fd5d8ad5605f1d3e98d8a8866a5d56c3e96afd64b%2Findex.html%3FisPreview%3Dtrue)

```
<html><body><h1>504 Gateway Time-out</h1>
The server didn't respond in time.
</body></html>

```

_If you don't have enough information above to solve the error by yourself (or to understand to which web service the error is related to, if any), please [file an issue](https://github.com/tobie/pr-preview/issues/new?title=Error%20not%20surfaced%20properly&body=See%20w3c/accname%23159.)._
</details>


jnurthen:
We have to think what the purpose of main and stable is. In ARIA we use 
main - this is the editors draft and contains features that don't have enough implementation commitment to put them into stable and into a working draft.
stable - stuff only ends up here when we have implementation commitment and issues filed to create authoring guidance.

What does this process look like for ACCNAME? If there is no difference and stuff only ends up in main once implementation commitment has been gathered (and tests written etc) then it is probably fine to simply merge main to stable - but if that isn't the same and there are features in the editors draft that aren't ready for stable then we probably should cherry pick.

There are also a bunch of commits in this PR from bots and changes in files which shouldn't be changed (all of the common directory) - some of which seem to undo changes from common.

MelSumner:
@jnurthen for the first concern, the work we've done until now is simply updating the spec to better reflect reality; we thought we would go for some changes after doing that kind of maintenance. So in that regard, I think it would be okay. 

However the second part of your comment makes me think that this approach for this occasion would not be okay.

PR #160, opened 2022-05-17
Specprod options

michael-n-cooper:
turn off optional checks when publishing editors' drafts
turn on automatic WD publication


MelSumner (approved):
I am not 100% about this syntax (just not familiar enough) but I approve the idea in theory. Is there someone else that can review the syntax just to confirm?

PR #165, opened 2022-08-12
link to definition of whitespace

scottaohara:
A similar update will happen in the ARIA spec, relating to https://github.com/w3c/core-aam/issues/128

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/165.html" title="Last updated on Oct 31, 2022, 6:09 PM UTC (bd38768)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/165/3f39e27...bd38768.html" title="Last updated on Oct 31, 2022, 6:09 PM UTC (bd38768)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -357,7 +357,7 @@
                 <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
               </details></div>
             </li>
-            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> [=attribute=] whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
+            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> [=attribute=] whose value is not the empty string, nor, when trimmed of [=ascii whitespace|whitespace=], is not the empty string:
               <ul>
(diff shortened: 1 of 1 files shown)


accdc (approved):
It sounds clear to me.

index.html
```diff
@@ -357,7 +357,7 @@ <h4>Computation steps</h4>
                 <pre class="example highlight"><code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code></pre>
               </details></div>
             </li>
-            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> [=attribute=] whose value is not the empty string, nor, when trimmed of white space, is not the empty string:
+            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> [=attribute=] whose value is not the empty string, nor, when trimmed of [=ascii whitespace|whitespace=], is not the empty string:
```
scottaohara, line 360: Per the suggestion to update core aam's reference for aria-roledescription to explicitly mention undefined and the empty string separately, does it make sense to also add the following change here?

If not in this PR, in a follow on?
```suggestion
            <li id="step2D">Otherwise, if computing a <a class="termref" data-lt="accessible name">name</a>, and if the <code>current node</code> has an <code>aria-label</code> [=attribute=] whose value is not undefined, not the empty string, nor, when trimmed of [=ascii whitespace|whitespace=], is not the empty string:
```

MelSumner: I think it makes sense to also add them here.

MelSumner (approved):
Thank you for making these changes!

Issue #93, opened 2020-09-04
How is the accessible name determined when an element contains a slot?


geoffrich:
How is the accessible name determined for elements that contain slots? For example, look at the following HTML which defines a custom button element containing a slot.

```html
<template id="button-template">
    <button><slot></slot></button>
</template>
<custom-button>Submit</custom-button>
<script>
    customElements.define(
    'custom-button',
    class extends HTMLElement {
        constructor() {
        super();
        const template = document.getElementById('button-template')
            .content;
        const shadowRoot = this.attachShadow({ mode: 'open' }).appendChild(
            template.cloneNode(true)
        );
        }
    }
    );
</script>
```

When inspecting the `<button>` element in the shadow root, both Chrome and Firefox say the accessible name is "Submit". However, I can't find anything in [the accessible name mapping spec](https://www.w3.org/TR/accname-1.1/#mapping_additional_nd_name) mentioning slots. Per step F.iii, each child node of the button would be recursively examined to determine the name. So, the text alternative for the slot would need to be determined. The slot does not have any childNodes, so the process stops there. It seems like assignedNodes would also need to be examined to compute the accessible name.

```js
const slot = document.querySelector('custom-button').shadowRoot.querySelector('slot')
console.log(slot.childNodes)        // NodeList []
console.log(slot.assignedNodes())   // Array [ #text "Custom name" ]
```

- When the spec says "For each child node of the current node," does it include assigned nodes when slots are being examined? 
- Should it be updated to mention assigned nodes or am I misunderstanding the spec? 

The browser implementations seem to be looking at assigned nodes as well as child nodes, so I'm wondering why the spec doesn't mention them.

jnurthen:
This doesn't sound like something accname should be defining. This should be defined in HTML-AAM IMO

jnurthen:
Accessible name for a button is defined at https://www.w3.org/TR/html-aam-1.0/#button-element-accessible-name-computation
As there is no aria-label or aria-labelledby on the button Accname is not involved. Hence closing this. If html-aam does not correctly answer the question then I encourage you to open an issue there.

JAWS-test:
@jnurthen 

> As there is no aria-label or aria-labelledby on the button Accname is not involved

Unfortunately I do not understand the argument. Accname is the general specification for determining names, not only in relation to aria-label and aria-labelledby. If custom elements (general, not only as buttons) are a special case concerning the determination of names, this should be considered in Accname. 

[...]

PR #167, opened 2022-09-15
Add steps for shadow roots and slots

### LINKED ISSUE #20: TAC and #presentation should cover shadow-DOM descendants (e.g. <video controls> in some user agents)
https://www.w3.org/WAI/ARIA/track/actions/1381

### LINKED ISSUE #51: Clarify if accessible names be calculated based on ancestor/descendant nested shadow DOM
It would be good to know if references such as aria-labelledby used in accessible name calc can cross into ancestor or descendant shadow DOM.  I agree that acc name can not be calculated outside of a shadow DOM -- but impact on nested shadow DOM is not clear.

nolanlawson:
Fixes #20 and #93 (related: #51).

Clarifies how accessible names are calculated for shadow roots and slots.

I believe this is accurate in terms of what UAs actually do to calculate the accessible name inside of shadow roots and slots, although WPT tests are probably needed.

# Implementation

* WPT tests: [PR](https://github.com/web-platform-tests/wpt/pull/36541), [landed - wpt.fyi](https://wpt.fyi/results/accname/name/shadowdom?label=experimental&label=master&aligned)
* Implementations:
   * WebKit: n/a
   * Gecko:  [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1858046)
   * Blink: [bug](https://issues.chromium.org/issues/40872282), landed
* Related HTML-AAM change: [issue](https://github.com/w3c/html-aam/issues/440)

***
<a href="https://pr-preview.s3.amazonaws.com/nolanlawson/accname/pull/167.html" title="Last updated on Feb 20, 2024, 6:38 PM UTC (493b13b)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/167/1e6661b...nolanlawson:493b13b.html" title="Last updated on Feb 20, 2024, 6:38 PM UTC (493b13b)">Diff</a>


nolanlawson:
Just chatted with @spectranaut –  the suggestion was that "displayed child nodes" is probably the wrong phrase here, and maybe "rendered child nodes" is more accurate. I'm happy with either one; naming things is hard. 🙂

jnurthen:
@jcsteh can you please look at this

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
                     <li>For <code>:after</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr>  textual content, without a space, to the textual content of the <code>current node</code>. </li>
                   </ul>
                 </li>
-                <li id="step2F.iii">For each child node of the <code>current node</code>:
+                <li id="step2F.iii">
+                  Determine the <code>displayed child nodes</code> of the <code>current node</code>:
```
cookiecrook, line 400:
```suggestion
                <li id="step2F.iii">
                  Determine the <code>rendered child nodes</code> of the <code>current node</code>:
```

index.html
```diff
@@ -224,6 +224,8 @@ <h4>Terminology</h4>
         <dd>The <abbr title="Document Object Model">DOM</abbr> [=nodes|node=] or [=element=] for which the text alternative is sought.</dd>
         <dt>Current node</dt>
         <dd>The <abbr title="Document Object Model">DOM</abbr> [=nodes|node=] currently traversed to compute the <code>root node</code>'s text equivalent.  Initially, the <code>current node</code> is the <code>root node</code>, but at later stages is either some descendant of the <code>root node</code>, or another referenced node.</dd>
+        <dt>Displayed child nodes</dt>
+        <dd>The <abbr title="Document Object Model">DOM</abbr> [=nodes=] that are rendered as child nodes of a given node when taking [=shadow root|shadow roots=] and [=slot|slots=] into consideration.</dd>
```
cookiecrook, line 228:
```suggestion
        <dt>Rendered child nodes</dt>
        <dd>The [=nodes=] that are rendered as child nodes of a given node when taking [=shadow root|shadow roots=] and [=slot|slots=] into consideration.</dd>
```

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
                 </li>
-                <li id="step2F.iii">For each child node of the <code>current node</code>:
+                <li id="step2F.iii">
+                  Determine the <code>displayed child nodes</code> of the <code>current node</code>:
                   <ol>
-                    <li id="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
-                    <li id="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with step 2.  Set the <code>result</code> to that text alternative.</li>
-                    <li id="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="step2F.iii.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>displayed child nodes</code> to be the child nodes of the [=shadow root=].</li>
+                    <li id="step2F.iii.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>displayed child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
```
cookiecrook, line 403:
```suggestion
                    <li id="step2F.iii.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>rendered child nodes</code> to be the child nodes of the [=shadow root=].</li>
                    <li id="step2F.iii.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>rendered child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
```

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
+                    <div><details>
+                      <summary>Comment:</summary>
+                      <p>A [=slot=] with [=slot/assigned nodes=] displays those nodes; otherwise it displays its default content, i.e. its child nodes.</p>
```
cookiecrook, line 406: I think "displays those nodes" may be ambiguous here. Which meaning of "displays" and also which nodes?

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
+                    </details></div>
+                    <li id="step2F.iii.c">Otherwise, set the <code>displayed child nodes</code> to be the child nodes of the <code>current node</code>.</li>
```
cookiecrook, line 408:
```suggestion
                    <li id="step2F.iii.c">Otherwise, set the <code>rendered child nodes</code> to be the child nodes of the <code>current node</code>.</li>
```

index.html
cookiecrook: "set" may also be ambiguous? Is it possible someone may misinterpret this means promoting the shadow root to be in the same doc root? I doubt it, but trying to anticipate alternate readings.

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
+                  </ol>
+                </li>
+                <li id="step2F.iiii">For each <code>displayed child node</code> of the <code>displayed child nodes</code> of the <code>current node</code>:
```
cookiecrook, line 411:
```suggestion
                <li id="step2F.iiii">For each <code>rendered child node</code> of the <code>current node</code>:
```

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
+                  <ol>
+                    <li id="step2F.iiii.a">Set the <code>current node</code> to the <code>displayed child node</code>.</li>
```
cookiecrook, line 413:
```suggestion
                    <li id="step2F.iiii.a">Set the <code>current node</code> to the <code>rendered child node</code>.</li>
```

cookiecrook (changes requested):
Added some diffs so that we can re-review with @spectranaut's suggestion. 

I'm not sure this addresses all of #51 since the submitter asked about aria-labelledby, which would presumably use a static element array when pointing into the shadow root. This is a great start, but an explicit mention of that new technique needed near the labelledby stage of the Name Computation?

index.html
nolanlawson: I'm using "set" in the sense of "set the variable named `renderedChildNodes` to be this value." Elsewhere in the spec it seems to use "set" in the sense of setting a variable as well.

Is there a better way I could factor this to be less ambiguous? I'm a bit unfamiliar with spec writing. 🙂

index.html
nolanlawson: I probably should have said "renders" instead of "displays."

The spec seems to purely speak in terms of the slot's [assigned nodes](https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element), remarking that:

> Returns slot's [assigned nodes](https://dom.spec.whatwg.org/#slot-assigned-nodes), if any, and slot's children otherwise

I can change the language to match this more closely.

nolanlawson: I think I found [the right spec language](https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element):

> A [`slot`](https://html.spec.whatwg.org/multipage/scripting.html#the-slot-element) element [represents](https://html.spec.whatwg.org/multipage/dom.html#represents) its [assigned nodes](https://dom.spec.whatwg.org/#slot-assigned-nodes), if any, and its contents otherwise.

I can pretty much just put this verbatim into this comment.

nolanlawson:
> I'm not sure this addresses all of https://github.com/w3c/accname/issues/51 since the submitter asked about aria-labelledby, which would presumably use a static element array when pointing into the shadow root.

The current `accname` spec says:

> if computing a [name](https://w3c.github.io/accname/#dfn-accessible-name), and the current node has an aria-labelledby [attribute](https://dom.spec.whatwg.org/#concept-attribute) that contains at least one valid IDREF […]

I think potentially there is nothing `accname` needs to do here. If an IDREF crosses a shadow boundary, then by definition it is not valid.

Perhaps this should go into the definition of an IDREF? I see that [in the aria spec](https://w3c.github.io/aria/#valuetype_idref), it says:

> **ID reference**
> Reference to the ID of another [element](https://dom.spec.whatwg.org/#concept-element) in the same document

It seems to me that this should read "in the same document _or shadow root_."

nolanlawson:
OTOH there is the related question of what happens if a `<slot>` element itself has an `aria-labelledby` or `aria-describeddby` (or `aria-label`, for that matter).

Whipping up [a quick CodePen](https://codepen.io/nolanlawson-the-selector/pen/eYrMBWj), and looking at the Chrome/Firefox Accessibility DevTools and VoiceOver+Safari, it looks like:

- Firefox ignores the `aria-labelledby`
- Chrome and Safari consider it a valid label

This _feels_ like a bug to me (in which case, Firefox's behavior is correct and the language in my PR is correct). But I admit I'm not sure.

nolanlawson:
> which would presumably use a static element array when pointing into the shadow root

Just grokked this point. For IDREF reflection (https://github.com/whatwg/html/pull/7934) – i.e. `ariaLabelledByElements`, `ariaDescribedByElements`, and friends – it looks like yes, we need to add explicit steps for that.

I think potentially this could be handled in a separate PR, since it's not strictly about shadow DOM. E.g. with `el1.ariaLabelledByElements = [el2]`, both `el1` and `el2` could be in light DOM.

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
                   <ol>
-                    <li id="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
-                    <li id="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with step 2.  Set the <code>result</code> to that text alternative.</li>
-                    <li id="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="step2F.iii.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>rendered child nodes</code> to be the child nodes of the [=shadow root=].</li>
+                    <li id="step2F.iii.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>rendered child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
+                    <div><details>
+                      <summary>Comment:</summary>
+                      <p>A [=slot=] element represents its [=slot/assigned nodes=], if any, and its contents (child nodes) otherwise.</p>
```
spectranaut, line 406: I think this comment confuses me a bit?

In the step you wrote: if the `current node` is a slot, then set the `rendered` nodes to it's `assigned nodes`, if any.

But in the note, it sounds like the slot might have `child nodes` separate from it's `assigned nodes`.  So in this algorithm, if there are no `assigned nodes` but there are `child nodes` of a slot -- the `child nodes` are not added to the `rendered child node` list.

Maybe the note is unnecessary and the step needs to include what to do with slots without assigned nodes?

Let me know if I'm misunderstanding something, I'm new to thinking about web components.

spectranaut (changes requested):
Hi Nolan, your right this change should have WPT tests, as that is how we currently keep a list of tests for accname. Here is an example test: https://github.com/web-platform-tests/wpt/blob/master/accname/name_text-title-manual.html

Let me know if anything is not straightforward, but you might be able to get the gist of the test by looking at it and others. These test what the browser exposes in the native API, which you can see in the `ATTAcomm` object. You would just want to make a new file, new title, some example html (give id="test" to the element you are testing the accname of), update the expected accname string for all APIs in the ATTAcomm `json`.

If you write them I'll review them! :)

spectranaut:
Also can you confirm, this algorithm represents what Safari and Firefox already do, but Chrome does not?

And here is the chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1374358

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
+                    </details></div>
+                    <li id="step2F.iii.c">Otherwise, set the <code>rendered child nodes</code> to be the child nodes of the <code>current node</code>.</li>
+                  </ol>
+                </li>
+                <li id="step2F.iiii">For each <code>rendered child node</code> of the <code>current node</code>:
```
jcsteh, line 411: nit: iiii should be iv here (and below), and iv below should be v.

index.html
nolanlawson: @jcsteh is correct; that's exactly what I was trying to communicate. [Here is a small CodePen](https://codepen.io/nolanlawson-the-selector/pen/rNvEZMX) to demonstrate.

Note that having _any_ nodes assigned to the `<slot>` (including an empty space) results in the assigned nodes being rendered. Otherwise, the child nodes (aka the default content) are rendered.

Perhaps there is a better way I can phrase it in the comment?

nolanlawson:
@spectranaut 

> Also can you confirm, this algorithm represents what Safari and Firefox already do, but Chrome does not?
> And here is the chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1374358

Correct, based on my testing (https://github.com/w3c/accname/issues/173), Safari and Firefox ignore the `aria-label` on the `<slot>`, whereas Chrome does not.

Based on https://github.com/w3c/html-aam/issues/440 though, it seems like we want to cover "elements that cannot be named" (e.g. `<slot>`) outside of the `accname` spec, in the `html-aam` spec instead. So I didn't handle it in this PR.

FWIW I did also open a PR to add some basic `accname` tests for shadow DOM and slots: https://github.com/web-platform-tests/wpt/pull/36541 . It broadly covers what this PR covers (although not `aria-description`).

index.html
```diff
@@ -394,11 +396,23 @@ <h4>Computation steps</h4>
...
                 </li>
-                <li id="step2F.iii">For each child node of the <code>current node</code>:
+                <li id="step2F.iii">
+                  Determine the <code>rendered child nodes</code> of the <code>current node</code>:
                   <ol>
-                    <li id="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
-                    <li id="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with step 2.  Set the <code>result</code> to that text alternative.</li>
-                    <li id="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="step2F.iii.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>rendered child nodes</code> to be the child nodes of the [=shadow root=].</li>
+                    <li id="step2F.iii.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>rendered child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
```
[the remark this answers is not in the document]
nolanlawson: My intention is to tweak the children-traversal part of the algorithm to traverse into slot assigned nodes rather than child nodes. The slot itself is not excluded from processing; you're correct.

Based on https://github.com/w3c/html-aam/issues/440 it looks like this should be handled in the `html-aam` spec instead?

rniwa:
As discussed during AOM meeting, it's important that there is no JS API that exposes the computed accessibility names that emanate out of a shadow DOM since that would violate shadow DOM's encapsulation. And so far we've concluded there is none so we're good here.

index.html
spectranaut: I see -- tbh now I'm just not so sure a note is needed at all?

index.html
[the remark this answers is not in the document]
jcsteh: > The slot itself is not excluded from processing; you're correct.

If I understand correctly, that means the Chromium bug isn't dealt with by this addition to the spec, correct? I'm not saying this isn't necessary, just clarifying for sure that it doesn't address the Chromium bug.

> it looks like this should be handled in the html-aam spec instead?

This gets back to my comment about tree traversal. As I see it, it's less to do with the fact that a slot can't be named and more to do with the fact that the slot itself (but not its rendered children) should be skipped in the traversal in the first place. I realise this might sound like nitpicking, but I think it's important in understanding how browsers actually do this.

index.html
nolanlawson: Makes sense to me! I removed the comment.

index.html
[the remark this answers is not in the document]
jcsteh: Oh blah. You're right: it's probably the fact that it has display: contents that excludes it from the Firefox a11y tree. As your test case shows, it does get exposed with display: block (because block elements have some semantic value by virtue of being block). That means that `<slot aria-label="foo">` probably would get exposed in Firefox except for [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1494196).

So it does look like we'll need some explicit provision that slots shouldn't be exposed. I'm not sure if them not having a name is enough. That said, this is a bit tricky because `<slot style="display: block;" tabindex="0">` does become focusable.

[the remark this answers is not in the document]
nolanlawson: Thanks, that makes sense! To identify this bug, we could try adding WPT tests that set `display: block` on the `<slot>`.

> That said, this is a bit tricky because `<slot style="display: block;" tabindex="0">` does become focusable.

Well, maybe `<slot>` _should_ be an element that can support naming? Unless https://github.com/w3c/accname/issues/173#issuecomment-1279626528 is the last word on this.

Note that I don't have a strong opinion; I'm just trying to determine the correct behavior.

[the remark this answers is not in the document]
scottaohara: In the case someone does `<slot style="display: block;" tabindex="0">`, we've gotten quite into the territory of author error correction.

If properly used, the slot element, along with elements like title, meta, template, etc... none of those should ever actually be exposed to a user and they should not be nameable.  But, if you force these elements to be rendered by changing their CSS display property, then they essentially become 'generic' elements.

I think it makes sense to call out that if an author forces elements like these to be exposed to the a11y tree, then browsers should handle them like generics.  But, again, I would stress this is correcting for author misuse and otherwise - if used properly - they cannot be named.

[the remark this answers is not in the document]
nolanlawson: > I think it makes sense to call out that if an author forces elements like these to be exposed to the a11y tree, then browsers should handle them like generics.

@scottaohara Thanks for the feedback! Do you think this should be done in `accname` or `html-aam`? I'm wondering if it makes more sense to put it in `html-aam` (per https://github.com/w3c/html-aam/issues/440), since in `accname` we're merely defining the tree traversal algorithm w.r.t `<slot>`s and their assigned nodes, not whether `<slot>`s can be named.

nolanlawson:
I did another read-through of the comments. I think most of the feedback has been addressed, and the PR on WPT should cover this PR: https://github.com/web-platform-tests/wpt/pull/36541

`ariaLabelledByElements`/`ariaDescribedByElements` is trickier, and is being handled in #170. There is also the related question of `ElementInternals` (https://github.com/w3c/core-aam/issues/152) and whether `<slot>`s can be named (https://github.com/w3c/html-aam/issues/440). But I don't believe these need to be solved to solve traversal for shadow roots / `<slot>`s (this PR).

index.html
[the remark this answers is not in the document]
scottaohara: yes, i think this should be handled in html aam per the linked issue.  i don't think accName should need to call this particular element out in any specific way. that's is what html aam can do.

jnurthen:
@MelSumner @accdc Does this need anything else to get this merged? It seems like all the review comments have been addressed.

accdc:
My vote is to merge. I'll wait for Melanie to confirm and then we are good to go.

cookiecrook (approved):
This looks okay to me now. Sorry I didn't get back to it after my initial review, but I agree my feedback has been addressed.

Of note, @nolanlawson, this will be testable in wpt/accname/name/comp_name_from_content.html once [WPT #39604](https://github.com/web-platform-tests/wpt/pull/39604/) lands. Please update the ReadMe in that same dir if you're the one making the change.

nolanlawson:
@cookiecrook No worries, thank you for the review!

jnurthen:
@spectranaut are your review comments addressed?

spectranaut (approved):
looks good to me!

index.html
```diff
@@ -437,14 +439,21 @@ <h4>Computation steps</h4>
...
+                <li id="comp_name_from_content_find_child" name="step2F.iii"><em>Determine Child Nodes:</em> Determine the <code>rendered child nodes</code> of the <code>current node</code>:
                   <ol>
-                    <li id="comp_name_from_content_for_each_child_set_current" name="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
-                    <li id="comp_name_from_content_for_each_child_recursion" name="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with the overall <a href="#comp_computation">Computation</a> step. Set the <code>result</code> to that text alternative.</li>
-                    <li id="comp_for_each_child_append" name="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="comp_name_from_content_find_child_has_shadow_root" name="step2F.iii.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>rendered child nodes</code> to be the child nodes of the [=shadow root=].</li>
```
cookiecrook, line 444: Repurposing the "step2F.iii.a" here breaks the old permalinks. We kept those so that legacy deep links in would not break, but the legacy `name` value should stay with the original reference (comp_name_from_content_for_each_child_set_current) and no new numbered `name` values should be added. 

New permalinks should used uniquely names `id` values only.

index.html
```diff
@@ -437,14 +439,21 @@ <h4>Computation steps</h4>
...
+                    <li id="comp_name_from_content_find_child_is_slot" name="step2F.iii.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>rendered child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
+                    <li id="comp_name_from_content_find_child_otherwise" name="step2F.iii.c">Otherwise, set the <code>rendered child nodes</code> to be the child nodes of the <code>current node</code>.</li>
                   </ol>
                 </li>
-                <li id="comp_name_from_content_return" name="step2F.iv">Return the <code>accumulated text</code> if it is not the empty string ("").</li>
+                <li id="comp_name_from_content_for_each_child" name="step2F.iv"><em>Name From Each Child:</em> For each <code>rendered child node</code> of the <code>current node</code>:
+                  <ol>
+                    <li id="comp_name_from_content_for_each_child_set_current" name="step2F.iv.a">Set the <code>current node</code> to the <code>rendered child node</code>.</li>
```
cookiecrook, line 451:
```suggestion
                    <li id="comp_name_from_content_for_each_child_set_current" name="step2F.iii.a"><!-- Leave legacy "step2F.iii.a" name value with this step, and don't add new name values. -->Set the <code>current node</code> to the <code>rendered child node</code>.</li>
```

Perhaps we should even add comments like above so this is less likely to happen in the future

cookiecrook (changes requested):
Need to move the name values back to their original steps. New numbered `name` values should not be added; only named `id` values like `id="comp_something_logical_and_unique_to_this_step_even_if_it_is_moved"`.

index.html
nolanlawson: No problem. I changed it so that:

1. New content has its own convention for the `name`
2. Pre-existing content (`comp_name_from_content_for_each_child`) has the same `name` as before

GitHub doesn't do a great job with the diff, but permalinks should not be broken anymore.

index.html
nolanlawson: Sounds good. I did add a comment just to be safe:

https://github.com/w3c/accname/blob/a9cfa2e946b0b4516fd8b32ba54d763f391f55df/index.html#L442

index.html
```diff
@@ -437,9 +439,17 @@ <h4>Computation steps</h4>
                     <li id="comp_name_from_content_pseudo_element_after">For <code>::after</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space, to the textual content of the <code>current node</code>. </li>
                   </ol>
                 </li>
-                <li id="comp_name_from_content_for_each_child" name="step2F.iii"><em>Name From Each Child:</em> For each child node of the <code>current node</code>:
+                <!-- The following section uses its own convention for the name attribute to avoid breaking existing permalinks for other sections-->
+                <li id="comp_name_from_content_find_child" name="stepFindChild"><em>Determine Child Nodes:</em> Determine the <code>rendered child nodes</code> of the <code>current node</code>:
```
cookiecrook, line 443:
```suggestion
                <li id="comp_name_from_content_find_child"><em>Determine Child Nodes:</em> Determine the <code>rendered child nodes</code> of the <code>current node</code>:
```

cookiecrook (approved):
Substantive change looks good, but editorial nits included.

index.html
```diff
@@ -437,9 +439,17 @@ <h4>Computation steps</h4>
...
                   <ol>
-                    <li id="comp_name_from_content_for_each_child_set_current" name="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
+                    <li id="comp_name_from_content_find_child_has_shadow_root" name="stepFindChild.a">If the <code>current node</code> has an attached [=shadow root=], set the <code>rendered child nodes</code> to be the child nodes of the [=shadow root=].</li>
+                    <li id="comp_name_from_content_find_child_is_slot" name="stepFindChild.b">Otherwise, if the <code>current node</code> is a [=slot=] with [=slot/assigned nodes=], set the <code>rendered child nodes</code> to be the [=slot/assigned nodes=] of the <code>current node</code>.</li>
+                    <li id="comp_name_from_content_find_child_otherwise" name="stepFindChild.c">Otherwise, set the <code>rendered child nodes</code> to be the child nodes of the <code>current node</code>.</li>
```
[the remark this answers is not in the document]
nolanlawson: OK, removed the new `name`s!

Jym77:
(semi-randomly finding that now)
Isn't this trying to redefine the [flat tree](https://drafts.csswg.org/css-scoping/#flat-tree) of CSS?
Shouldn't we just change step 2.F.iii from
> For each child node of the current node

to
> For each child node of the current node **in the flat tree**

that should already pretty much take care of all the shadow DOM magic 🤔

nolanlawson:
@Jym77 I wasn't even aware that there was a concept of a "flattened tree." The link you provide is from the CSS spec – if there's something in the HTML spec, maybe we could reference it here to simplify the logic?

nolanlawson:
BTW the WPT tests for this PR have already been merged: https://github.com/web-platform-tests/wpt/pull/36541

Maybe we can merge and then refactor later as necessary?

jcsteh:
Realistically, accessibility implementations depend on layout rendering to some extent; e.g. display: none, visibility: hidden. Given that, it's probably not unreasonable to refer to a CSS concept here. FWIW, Gecko's accessibility engine internally uses the flat tree to build the accessibility tree and I assume other browsers might do similar.

alice:
> Gecko's accessibility engine internally uses the flat tree to build the accessibility tree and I assume other browsers might do similar.

Blink does the same (technically the "layout tree builder traversal" which includes pseudo-elements, but otherwise just uses the flat tree traversal), since the whole `display: contents` [incident](https://bugs.chromium.org/p/chromium/issues/detail?id=835455). (Previously it walked the layout tree directly, which mostly amounts to the same thing anyway.)

MelSumner:
I am not sure what we want to do about this so I'm going to mark it for the agenda, and read through it all in the meantime.

spectranaut:
This looks read to land, @jnurthen will resolve conflicts

PR #168, opened 2022-09-20
Fixed handling of surrounding whitespace for CSS pseudo elements, inline and block level elements, and embedded widgets.

accdc:
This change adds logic to handle block level and inline elements where it is necessary to add a space at the beginning or end of accumulated text in order to return the most accessible names and descriptions for the accessibility tree.

Before merging, we need to confirm:
- [ ] Test coverage (add link to tests)
- Need updates to browsers?
    - [ ] Chrome: yes/no
    - [ ] Safari: yes/no
    - [ ] Firefox: yes/no

[Update: assuming this stays as a note only, there will be no WPT or implementation issues... Those would come with #225. –@cookiecrook]

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/168.html" title="Last updated on Jan 19, 2024, 2:43 AM UTC (8ecf859)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/168/a00a582...8ecf859.html" title="Last updated on Jan 19, 2024, 2:43 AM UTC (8ecf859)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -490,6 +490,9 @@
               <div class="note">
                 <p>This step can apply to the child nodes themselves, which means the computation is recursive and results in text collected from all the elements in the <code>current node</code>'s subtree, no matter how deep it is. However, any given descendant [=nodes|node's=] text alternative can result from higher precedent markup described in steps B through D above, where "Namefrom: author" attributes provide the text alternative for the entire subtree. </p>
               </div>
+              <div class="note" id="note-concatenate-inline-contents-with-space">
+                <p>18 January 2024: The ARIA Working Group is considering the feasibility of joining text strings with and without spaces, depending on the CSS <code>display</code> value of the <code>current node</code>, and its adjacent nodes and pseudo-elements. The ongoing discussion is in <a href="https://github.com/w3c/accname/issues/225">AccName #225</a>.</p>
+              </div>
             </li>
             <li id="comp_text_node"><span id="step2G"><!-- Don't link to this legacy numbered ID. --></span><em>Text Node:</em> Otherwise, if the <code>current node</code> is a Text [=Node=], return its textual contents.</li>
             <li id="comp_recursive_name_from_content"><span id="step2H"><!-- Don't link to this legacy numbered ID. --></span><em>Recursive Name From Content:</em> Otherwise, if the <code>current node</code> is a descendant of an element whose <a class="termref">Accessible Name</a> or <a class="termref">Accessible Description</a> is being computed, and contains descendants, proceed to <a href="#comp_name_from_content_reset">Name From Content Reset</a>.</li>


accdc:
Note, as confirmed in Chrome Beta, this is meant to explicitly codify what is already happening in the accessibility tree.

Test pages:

Block level and inline elements: https://whatsock.github.io/w3c-alternative-text-computation/Proposed%20Name%20and%20Description%20Tests/Name%20file-label-inline-block-elements.html

Block level and inline styles: https://whatsock.github.io/w3c-alternative-text-computation/Proposed%20Name%20and%20Description%20Tests/Name%20file-label-inline-block-styles.html

index.html
```diff
@@ -390,15 +391,30 @@ <h4>Computation steps</h4>
...
                   <ul>
-                    <li>For <code>:before</code> pseudo elements, <a class="termref">User agents</a> MUST prepend <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space, to the textual content of the <code>current node</code>. </li>
-                    <li>For <code>:after</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr>  textual content, without a space, to the textual content of the <code>current node</code>. </li>
+                    <li>For <code>:before</code> pseudo elements, <a class="termref">User agents</a> MUST prepend <abbr title="Cascading Style Sheets">CSS</abbr> textual content to the textual content of the <code>current node</code> as follows:
+                    <ul>
+                      <li>If the pseudo element includes spacing between it and the textual content of the <code>current node</code>, <a class="termref">User agents</a> MUST prepend <abbr title="Cascading Style Sheets">CSS</abbr> textual content, with a space following, to the textual content of the <code>current node</code>. </li>
```
cookiecrook, line 396: Emphasis mine:

> If the pseudo element includes _spacing_ between it and the textual content of the current node

This is ambiguous and potential clarifications may be difficult to implement or costly for performance. There's no "spacing" between an element and its pseudos, but the visual spacing is depending on many factors like padding, margin, position, text-indent, borders, position of background elements or list-style-images...

index.html
```diff
@@ -390,15 +391,30 @@ <h4>Computation steps</h4>
...
                   </ul>
                 </li>
                 <li id="step2F.iii">For each child node of the <code>current node</code>:
                   <ol>
                     <li id="step2F.iii.a">Set the <code>current node</code> to the child node.</li>
                     <li id="step2F.iii.b">Compute the text alternative of the  <code>current node</code> beginning with step 2.  Set the <code>result</code> to that text alternative.</li>
-                    <li id="step2F.iii.c">Append the <code>result</code> to the <code>accumulated text</code>. </li>
+                    <li id="step2F.iii.c">
+                    <ol>
+                      <li>If the <code>current node</code> includes spacing between adjacent elements and itself, add a space before and after the <code>result</code>.</li>
```
cookiecrook, line 414: ditto

index.html
```diff
@@ -390,15 +391,30 @@ <h4>Computation steps</h4>
...
+                      <li>Otherwise, <a class="termref">User agents</a> MUST prepend <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space following, to the textual content of the <code>current node</code>. </li>
+                    </ul>
+                    </li>
+                    <li>For <code>:after</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr> textual content to the textual content of the <code>current node</code> as follows:
+                    <ul>
+                      <li>If the pseudo element includes spacing between the textual content of the <code>current node</code> and itself, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr> textual content, with a space preceding, to the textual content of the <code>current node</code>. </li>
```
cookiecrook, line 402: ditto

cookiecrook (changes requested):
I don't have much confidence that we can specify an algorithm that will well work here. Perhaps better to choose a default (join w/ space for pseudos) and maybe make an explicit author opt-out?

cookiecrook:
Changed title b/c according to the diff, I think "block and inline styled elements" is just referring to pseudo elements. If this PR was associated with an issue that included a broader scope, please link it.

accdc:
FYI, Actually this PR does include changes to handle both inline and block level elements as well as CSS pseudo elements, since the only difference between has to do with how both are styled implicitly or explicitly. E.G. Styling a div as display:inline.

I don't mind if changes are needed to make this more understandable, but we do need something that says what the difference is, because at present there is nothing that says there is a difference.

Perhaps actually using the words inline or block level styling is appropriate, I'm not sure. Suggestions are welcome to make this work.

I'm not sure how you changed the title, but if you could please put that back to refer to both inline and block level styled elements I would appreciate it.

MelSumner (approved):
👍

index.html
MelSumner: wow! It seems like this term is definitely overloaded for some folks. 

What if we said "if the pseudo element includes a space between it and the textual content of the current node..."

Maybe we could provide a code sample that demonstrates what is meant here? 

I don't want folks to overthink it but it's easy to do that with an overloaded term.

cookiecrook: Trying to clarify more of the ambiguity.

# "spacing" or "a space" as in what?

## whitespace characters?
- "space [character] between" (e.g. " " or other whitespace)?: TMK, there is never a character _between_ an element and its pseudo.
- space character inside the element (e.g. first child is a text node and startswith " ") "between" the visible element text and the `::before` contents?: e.g. A prepended or appended space character inside the element `<span> foo </span>? This is trimmed by other aspects of the algorithm.
- space character inside the element's pseudo?: e.g. A content value that begins with (for `::after`) or ends with (for `::before`) some whitespace character?

## Other non-whitespace-character space?
- left padding or margin of the element that has a `::before` psuedo?
- right padding or margin of the element that has an `::after` psuedo?
- left padding or margin `::after` pseudo itself?
- right padding or margin `::before` pseudo itself?
- invisible borders (e.g. `border-color: rgba(256,256,256,0);`) on either the element or pseudo?
- a pseudo with a fixed width whose background image does not fill the entire width (such as and image bullet in a bulleted list)

I'm certain this list is not exhaustive, but hopeful it shows the algorithm complexity that's being requested.

cookiecrook: I think @aleventhal would agree, but I'd like his opinion, too.

cookiecrook: Or @cyns?

jnurthen:
This seems to touch some of the same text as #183 - need to resolve those changes in combination with that PR once it is merged.

MelSumner:
ok I think I have resolved the merge conflicts. @cookiecrook do you feel like your comments are resolved or do you want to take another pass?

jnurthen (dismissed):
This appears to have discarded all the changes I carefully merged into main for the new idref and format for each item. Please restore all the idrefs and the visible name for each step.

MelSumner:
> This appears to have discarded all the changes I carefully merged into main for the new idref and format for each item. Please restore all the idrefs and the visible name for each step.

@jnurthen i do see now that some of the idrefs that weren't in the merge conflict would have been removed. I only caught the ones in the merge conflict, sorry about that! I'll update. 

@accdc I'll work on this 👍

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
-                <li id="comp_embedded_control_range"><em>Range:</em> If the embedded control has role <a class="role-reference" href="#range">range</a> (e.g., a <a class="role-reference" href="#spinbutton">spinbutton</a> or <a class="role-reference" href="#slider">slider</a>):
-                  <ol>
-                    <li id="comp_embedded_control_range_valuetext">If the <code>aria-valuetext</code> property is present, return its value, </li>
-                    <li id="comp_embedded_control_range_valuenow">Otherwise, if the <code>aria-valuenow</code> property is present, return its value,</li>
-                    <li id="comp_embedded_control_range_host_language_value">Otherwise, use the value as specified by a host language attribute. </li>
-                  </ol>
+             <ul>
+                <li>If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
```
MelSumner, line 427:
```suggestion
                <li id="comp_embedded_control_textbox">If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
```

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
+                <li>If the embedded control has role <a class="role-reference" href="#combobox">combobox</a> or <a class="role-reference" href="#listbox">listbox</a>, return the text alternative of the chosen <a class="role-reference" href="#option">option</a>.</li>
```
MelSumner, line 428:
```suggestion
                <li id="comp_embedded_control_combobox_or_listbox">If the embedded control has role <a class="role-reference" href="#combobox">combobox</a> or <a class="role-reference" href="#listbox">listbox</a>, return the text alternative of the chosen <a class="role-reference" href="#option">option</a>.</li>
```

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
+                <li>If the embedded control has role <a class="role-reference" href="#range">range</a> (e.g., a <a class="role-reference" href="#spinbutton">spinbutton</a> or <a class="role-reference" href="#slider">slider</a>):
```
MelSumner, line 429:
```suggestion
                <li id="comp_embedded_control_range">If the embedded control has role <a class="role-reference" href="#range">range</a> (e.g., a <a class="role-reference" href="#spinbutton">spinbutton</a> or <a class="role-reference" href="#slider">slider</a>):
```

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
+                  <ul>
+                    <li>If the <code>aria-valuetext</code> property is present, return its value, </li>
```
MelSumner, line 431:
```suggestion
                    <li id="comp_embedded_control_range_valuetext">If the <code>aria-valuetext</code> property is present, return its value, </li>
```

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
+                    <li>Otherwise, if the <code>aria-valuenow</code> property is present, return its value,</li>
```
MelSumner, line 432:
```suggestion
                    <li id="comp_embedded_control_range_valuenow">Otherwise, if the <code>aria-valuenow</code> property is present, return its value,</li>
```

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
+                    <li>Otherwise, use the value as specified by a host language attribute. </li>
```
MelSumner, line 433:
```suggestion
                    <li id="comp_embedded_control_range_host_language_value">Otherwise, use the value as specified by a host language attribute. </li>
```

MelSumner:
ok @jnurthen I went through it again;  apologies for missing that top section. I think I caught them all this time.

accdc:
Thanks Melanie, you're awesome!

accdc:
For those assigned reviewers, we are ready to pull this in if you have no further objections.

MelSumner:
In order to keep this work moving forward, I'm going to move any comments or concerns to a separate issue, so we can merge this PR.

MelSumner:
@jnurthen could you look at this again before the holidays, if possible? Thank you!!

daniel-montalvo:
Hi @jnurthen @cookiecrook 

My understanding is that there have been changes in this PR after you both reviewed this. Are you planning on having a look  at this again?

Thank you

aleventhal:
Is this statement still true? “Note, as confirmed in Chrome Beta, this is meant to explicitly codify what is already happening in the accessibility tree.”

spectranaut:
Discussed on: https://www.w3.org/2024/01/04-aria-minutes.html#t05

cookiecrook:
> Discussed on: https://www.w3.org/2024/01/04-aria-minutes.html#t05

The TLDR is: the feedback re: "space" and "spacing" hasn't been changed since my last review, but I offered to try to propose some changes in a diff or commit to the branch. I recall the ARIA spec now links to the HTML definition of whitepace, so we may be able to reuse that to resolve the problems [mentioned above (comment)](https://github.com/w3c/accname/pull/168#discussion_r1037633578).

cookiecrook (changes requested):
I took an action to try to fix this, but the only correction I can think to make is to delete most of the diff. We may need to deep dive on this one, because the entire change relies on this concept:

> If the pseudo element includes spacing between it and the textual content of the current node

And that description of "spacing" is meaningless in the context of textual space as @accdc and @MelSumner alluded to on Thursday's call. There can never be _textual_ space between an element and its pseudo. 

There can be visual spacing like a margin, or interior whitespace character inside the pseudo's `content` string or the element's `innerText` but there are no characters between between an element in the DOM and a pseudo element that is not in the DOM. The layout proximity is composited but, as I understand the CSS generated content spec, there's no time when exterior textual space comes into this equation.

index.html
cookiecrook: @MelSumner 
> What if we said "if the pseudo element includes a space between it and the textual content of the current node..."

Do you mean this space?

```html
<div> text</div>
<!-- ^ -->
```

Or one of these spaces?

```css
  content: "text ";
  /*            ^ */
  content: "text " / "alt";
  /*            ^ */
  content: "text" / "alt ";
  /*                    ^ */
```

accdc:
This seems much more complicated than it needs to be.

For normal elements, if they are styled as block level elements, a space should be added to separate them from the prior and following content. Otherwise if styled as inline elements then no space should be added.

Why can't a pseudo element follow the same logic? It too can be either a block level or inline styled element.

I don't expect this to fix everything, but if we at least can get wording to handle this correctly it will handle most cases where these conditions are met.

accdc:
Just as a bit of background, the reason why all this wording about visual spacing and so is in the spec text, is because apparently we can't actually say "block level element". Basically that's all that boils down to.

So, can you think of how to say block level element without actually saying "block level element" within the spec?

I tried and clearly failed to do this in a way that makes sense.

spectranaut:
@accdc @MelSumner @cookiecrook -- Bryan asked for a deep dive in order to resolve these issues, but the agenda this week is light, so I'll put this item on the agenda for this week (Jan 18th) in the hopes that we can resolve it during normal meeting time. If someone can not attend this week we will move it, but if you can attend please come prepared to discuss.

MelSumner:
@spectranaut I'll make every effort to be there, I have an appt in the morning but I'll do my best. Thanks for adding it to the agenda!

cookiecrook:
WG meeting resolution: @cookiecrook to remove the problematic lines from the PR to move this forward to CR. Will file a new issue about possibly revisting CSS-AAM for display values et al.

spectranaut:
Minutes from today's meeting: https://www.w3.org/2024/01/18-aria-minutes.html#t05

index.html
```diff
@@ -423,15 +423,16 @@ <h4>Computation steps</h4>
...
-              <ol>
-                <li id="comp_embedded_control_textbox"><em>Textbox:</em> If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
-                <li id="comp_embedded_control_combobox_or_listbox"><em>Combobox/Listbox:</em> If the embedded control has role <a class="role-reference" href="#combobox">combobox</a> or <a class="role-reference" href="#listbox">listbox</a>, return the text alternative of the chosen <a class="role-reference" href="#option">option</a>.</li>
-                <li id="comp_embedded_control_range"><em>Range:</em> If the embedded control has role <a class="role-reference" href="#range">range</a> (e.g., a <a class="role-reference" href="#spinbutton">spinbutton</a> or <a class="role-reference" href="#slider">slider</a>):
-                  <ol>
+             <ul>
+                <li id="comp_embedded_control_textbox">If the embedded control has role <a class="role-reference" href="#textbox">textbox</a>, return its value.</li>
```
cookiecrook, line 427: Why were the step names removed? These were added when we switched off numbered IDs so that people could remember a step by its unified name. This PR seems unrelated, so I'm putting these back in.

index.html
```diff
@@ -470,18 +471,33 @@ <h4>Computation steps</h4>
...
-                  <ol>
-                    <li id="comp_name_from_content_pseudo_element_before">For <code>::before</code> pseudo elements, <a class="termref">User agents</a> MUST prepend <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space, to the textual content of the <code>current node</code>. </li>
-                    <li id="comp_name_from_content_pseudo_element_after">For <code>::after</code> pseudo elements, <a class="termref">User agents</a> MUST append <abbr title="Cascading Style Sheets">CSS</abbr> textual content, without a space, to the textual content of the <code>current node</code>. </li>
-                  </ol>
+                <li id="comp_name_from_content_reset"><span id="step2F.i"><!-- Don't link to this legacy numbered ID. --></span>Set the <code>accumulated text</code> to the empty string.</li>
```
cookiecrook, line 474: Same here and throughout. The step name was removed. Was this accidental?

index.html
cookiecrook: On the call, Bryan said this should refer to whether or not the element was an `inline` or similar value, versus a `block` or similar display value. I presume that would mean if either the adjacent elements (or their pseudo when joining pseudos) had a "block-like" display (perhaps any display value other than `inline`) then use a space joiner... 

Unknown at the moment whether that is implementable, and also unknown is how far the render tree should consider "adjacent" render objects... For example, if the adjacent DOM element is `display:none`, is the expectation that the renderer check the next adjacent element or pseudo that is not `display:none`? What about a `display:inline` element with `visibility:hidden`?

cookiecrook (changes requested):

index.html
cookiecrook: @MelSumner Is that related to this other merge problem? https://github.com/w3c/accname/pull/168#issuecomment-1800914469

cookiecrook: @Melsumner Nm, I see this just came off a much older branch, so these were left out of a manual merge. Hold off on changing anything. I'm going to reset the branch to main and then manually merge any of Bryan's original diff that needs to come back.

cookiecrook:
So that thread completeness isn't lost, here was the original substantive diff from Bryan: https://github.com/w3c/accname/pull/168/commits/fa737b5aacb84d7b8e50664d6b7b3f02fdcb85d5

cookiecrook:
@accdc I can't re-request review since this is your PR, but I've replaced the prior diffs with my own... Please review this note and the new issue #225. Thanks.

accdc:
Thank you!

 @MelSumner  can you confirm this looks good to you now as well?

If yes I think we are good to go.

MelSumner (approved):

PR #170, opened 2022-09-29
Add steps for `ariaLabelledByElements` and `ariaDescribedByElements`

nolanlawson:
Related to #167, and attempts to address #51.

With https://github.com/whatwg/html/pull/7934 we have the addition of the `ariaLabelledByElements` and `ariaDescribedByElements` properties, which reference [attr-associated elements](https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#attr-associated-elements) (a [FrozenArray](https://webidl.spec.whatwg.org/#idl-frozen-array)).

https://github.com/whatwg/html/pull/7934 already explains how these arrays interact with shadow DOM, so (AIUI) the `accname` spec does not need to explicitly recapitulate these steps. But for the record:

> Let elements be an empty [list](https://infra.spec.whatwg.org/#list).
>
> If element's [explicitly set attr-elements](https://html.spec.whatwg.org/#explicitly-set-attr-elements) is not null, then:
>
> [For each](https://infra.spec.whatwg.org/#list-iterate) attrElement in the element's [explicitly set attr-elements](https://html.spec.whatwg.org/#explicitly-set-attr-elements):
>
> If attrElement is not a [descendant](https://dom.spec.whatwg.org/#concept-tree-descendant) of any of element's [shadow-including ancestors](https://dom.spec.whatwg.org/#concept-shadow-including-ancestor), then [continue](https://infra.spec.whatwg.org/#iteration-continue).
> 
> [Append](https://infra.spec.whatwg.org/#list-append) attrElement to elements.

So I believe `accname` needs only to make mention of the `ariaLabelledByElements`/ `ariaDescribedByElements` arrays in the same places where it makes mention of `aria-labelledby` and `aria-describeddby`. If two elements are in shadow roots that cannot be associated (i.e. one is not a descendant of the other's shadow-including ancestors), then the array would not contain that element in the first place.

/cc @mrego

***
<a href="https://pr-preview.s3.amazonaws.com/nolanlawson/accname/pull/170.html" title="Last updated on Oct 6, 2022, 9:12 PM UTC (2409e26)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/170/bb65de3...nolanlawson:2409e26.html" title="Last updated on Oct 6, 2022, 9:12 PM UTC (2409e26)">Diff</a>

### DIFF
--- a/index.html
+++ b/index.html
@@ -308,13 +308,13 @@
             </li>
             <li id="step2B">Otherwise:
               <ul>
-                <li>if computing a <a class="termref" data-lt="accessible name">name</a>, and the <code>current node</code> has an <code>aria-labelledby</code> [=attribute=] that contains at least one valid IDREF or an <code>ariaLabelledByElements</code> array that contains at least one valid element, and the <code>current node</code> is not already part of an <code>aria-labelledby</code> traversal, process its IDREFs or elements in the order they occur: </li>
-                <li>or, if computing a <a class="termref" data-lt="accessible description">description</a>, and the <code>current node</code> has an <code>aria-describedby</code> [=attribute=] that contains at least one valid IDREF or an <code>ariaDescribedByElements</code> array that contains at least one valid element, and the <code>current node</code> is not already part of an <code>aria-describedby</code> traversal, process its IDREFs or elements in the order they occur:
+                <li>if computing a <a class="termref" data-lt="accessible name">name</a>, and the <code>current node</code> has an <code>ariaLabelledByElements</code> array property that contains at least one valid element, and the <code>current node</code> is not already part of an <code>aria-labelledby</code> traversal, process its <code>ariaLabelledByElements</code> in the order they occur: </li>
+                <li>or, if computing a <a class="termref" data-lt="accessible description">description</a>, and the <code>current node</code> has an <code>ariaDescribedByElements</code> array property that contains at least one valid element, and the <code>current node</code> is not already part of an <code>aria-describedby</code> traversal, process its <code>ariaDescribedByElements</code> elements in the order they occur:
                   <ol>
                     <li id="step2B.i">Set the <code>accumulated text</code> to the empty string.</li>
-                    <li id="step2B.ii">For each IDREF or element in the array:
+                    <li id="step2B.ii">For each element in the array property:
                       <ol>
-                        <li id="step2B.ii.a">Set the <code>current node</code> to the element referenced by the IDREF or element in the array.</li>
+                        <li id="step2B.ii.a">Set the <code>current node</code> to the element.</li>
                         <li id="step2B.ii.b">Compute the text alternative of the <code>current node</code> beginning with step 2. Set the <code>result</code> to that text alternative.</li>
                         <li id="step2B.ii.c">Append the <code>result</code>, with a space, to the <code>accumulated text</code>.</li>
                       </ol>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -308,13 +308,13 @@ <h4>Computation steps</h4>
...
+                <li>or, if computing a <a class="termref" data-lt="accessible description">description</a>, and the <code>current node</code> has an <code>aria-describedby</code> [=attribute=] that contains at least one valid IDREF or an <code>ariaDescribedByElements</code> array that contains at least one valid element, and the <code>current node</code> is not already part of an <code>aria-describedby</code> traversal, process its IDREFs or elements in the order they occur:
                   <ol>
                     <li id="step2B.i">Set the <code>accumulated text</code> to the empty string.</li>
-                    <li id="step2B.ii">For each IDREF:
+                    <li id="step2B.ii">For each IDREF or element in the array:
                       <ol>
-                        <li id="step2B.ii.a">Set the <code>current node</code> to the node referenced by the IDREF.</li>
+                        <li id="step2B.ii.a">Set the <code>current node</code> to the element referenced by the IDREF or element in the array.</li>
```
nolanlawson, line 317: Changed this from `node` to `element` since technically an IDREF can only point to an element, since only elements can have IDs.

nolanlawson:
After re-reading this discussion (https://github.com/whatwg/html/issues/8306), I think this PR may need to be amended.

It seems `aria-labelledby` and `aria-describedby` may get "out of sync" with their `FrozenArray` counterparts. For instance, the `ariaLabelledByElements` may have no IDs (in which case `aria-labelledby` is an empty string), or the `ariaLabelledByElements` may be moved into an invalid shadow root (in which case `aria-labelledby` will still have the "stale" ID, or possibly be the empty string if #8306 is fixed).

So perhaps we should change the `accname` spec to _only_ make reference to the `FrozenArray`s, since they are guaranteed to be correct. This would be quite a big change to the spec, though.

jnurthen:
I think this is going to need to change once #69 lands

nolanlawson:
> You cannot refer to the public API here. Instead you need to use and reference the internal concepts […]

I'm wondering if, at this point, it makes more sense to rewrite this spec from scratch to use the a11y tree rather than the DOM tree?

I've already gotten [feedback](https://github.com/w3c/accname/pull/167#pullrequestreview-1148595405) from @jsteh that Firefox walks the a11y tree rather than the DOM tree. If WebKit and Chromium do the same, then it would be a bit odd to have this spec describe something that browsers don't actually implement.

That said, I'm happy to proceed as-is to avoid making the perfect the enemy of the good.

cookiecrook:
> I'm wondering if, at this point, it makes more sense to rewrite this spec from scratch to use the a11y tree rather than the DOM tree?
> 
> I've already gotten https://github.com/w3c/accname/pull/167#pullrequestreview-1148595405 from @jsteh that Firefox walks the a11y tree rather than the DOM tree. If WebKit and Chromium do the same, then it would be a bit odd to have this spec describe something that browsers don't actually implement.

There isn’t a standard accessibility tree between platforms or engines, so that would be even more difficult. Each engine implements its own internal tree that is then vended to the platform accessibility API, to create the platform specific tree.

WebKit based its implementation off the render tree, Gecko off the DOM tree. Chromium moved a little closer to Firefox since the fork from WebKit, but each of the engine-specific accessibility trees are still unique to some degree.

annevk:
I think most browsers use some variant of the render/layout tree. E.g., they all hide `display:none` subtrees and such as far as I know.

Anyway, defining that is somewhat orthogonal to this particular issue as even if we did that, in order to get from an item in the render tree to its labelled elements, we'd have to poke at its corresponding DOM node and look at the internal concepts I mentioned above.

MelSumner:
Reading through this PR and all of the comments, I wonder if we have a path forward with this particular PR or maybe a different PR is needed? @nolanlawson WDYT?

nolanlawson:
@MelSumner Yes, I think that this PR probably needs to be rewritten from scratch at this point. In any place where we are referring to `aria-labelledby` or `aria-describedby`, we need to instead refer to the _attr_-associated elements.

PR #206, opened 2023-10-06
update accname 'flash screen' example to use native clickable label element

### LINKED ISSUE #65: Example 3 is wrong
https://w3c.github.io/accname/#example-3:

> `<div role="checkbox" aria-checked="false">Flash the screen <span role="textbox" aria-multiline="false"> 5 </span> times</div>`

According to https://www.w3.org/TR/wai-aria-1.1/#checkbox a checkbox has: "Children Presentational: True". Therefore the checkbox cannot contain an input field.

In my opinion, that would be correct: 

```
<div role="checkbox" aria-checked="false" aria-labelledby="label"></div>
<div id="label">Flash the screen <span role="textbox" aria-multiline="false"> 5 </span> times</div>
```

For the example to be correct, both interactive elements need a `tabindex=0` and the `textbox` a label. However, the following example would be simpler:

```
<div role="checkbox" tabindex="0" aria-checked="false" aria-labelledby="label"></div>
<label id="label">Flash the screen <input type="text" value="5"> times</label>
```

cookiecrook:
Closes #65

***
<a href="https://pr-preview.s3.amazonaws.com/w3c/accname/pull/206.html" title="Last updated on Nov 3, 2023, 10:39 PM UTC (c5b0566)">Preview</a> | <a href="https://pr-preview.s3.amazonaws.com/w3c/accname/206/114fe15...c5b0566.html" title="Last updated on Nov 3, 2023, 10:39 PM UTC (c5b0566)">Diff</a>


scottaohara:
@cookiecrook @MelSumner should we call out the fact that this example is invalid HTML?   i mean, it's far easier to use this pattern than the 'correctly nested' markup pattern, which could look like:
```
<label for="flash" id=label>
  <input type="checkbox" id="flash" aria-labelledby="label number times">
  Flash the screen 
</label>
<label for=number>
  <input type="text" value="5" aria-label="number of times" id=number> <span id=times>times</span>.
</label>
```

but a label is only allowed to contain a single labellable element, and if it weren't for the `for` attribute pointing to the checkbox, this would start breaking in other ways which wouldn't produce the described accName - but the use of `for` isn't referenced in the 'consider this' text as being necessary to make this work as the example says it will.

sorry, really don't mean to be so pedantic about this, but it does just strike me as odd for a spec to use an invalid example without at least mentioning it is invalid.  ☹️

accdc (approved):

scottaohara (changes requested):
just making this an official 'review' per my above comment.

again not in opposition to the example, i just think it needs to state it is invalid markup so as to not have this spec seemingly imply something is valid which html says is not.

cookiecrook:
I'm only slightly embarrassed to say I didn't think about the fact that two input in a label would make it invalid especially since the `for`/`id` relationship is explicitly defined. I don't think any accessibility implementations would be tripped up by the inclusion of the second input, which also has an explicitly defined aria-label. I agree we shouldn't use an invalid example though.

cookiecrook:
But I also don't like the replacement example. I don't think any one would use that awkward format. Considering other solutions that authors might use.

cookiecrook:
@scottaohara how one of these options

1. Avoid the second input in the label.
```html
<label for="flash">
  <input type="checkbox" id="flash">
  Flash the screen <span tabindex="0" role="textbox" aria-label="number of times" contenteditable>5</span> times.
</label>
```

2. Like yours, but avoid the labels entirely.
```html
  <input type="checkbox" id="flash" aria-labelledby="flash number times">
  <span id="flash">Flash the screen</span>
  <input id="number" type="text" value="5" aria-label="number of times">
  <span id="times">times.</span>
```

scottaohara:
fwiw, @cookiecrook, i agree - it 'should/could' to be "OK" - though i believe I've run into some windows based voice dictation software falling down on this sort of thing, where screen readers generally handle it ok these days.  anyway... appreciate your alternatives.

and yeah, i think either of those examples works.  I really appreciate the first one, specifically as a "html doesn't allow this?  oh yeah?"  it made me grin.

index.html
```diff
@@ -440,10 +440,11 @@ <h4>Computation steps</h4>
...
               <div>
                 <summary>Example:</summary>
-                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the user has entered "5" for the embedded <a class="role-reference" href="#textbox">textbox</a>, the complete label is "Flash the screen 5 times", e.g.: </p>
-                <pre class="example highlight">
-                  <code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code>
-                </pre>
+                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the input field has the value "5", the complete label is "Flash the screen 5 times", e.g.:</p>
+                <pre class="example highlight"><code>&lt;label for="flash"&gt;
+  &lt;input type="checkbox" id="flash"&gt;
+  Flash the screen &lt;input type="text" value="5" aria-label="number of times"&gt; times.
```
cookiecrook, line 446:
```suggestion
  Flash the screen &lt;span tabindex="0" role="textbox" aria-label="number of times" contenteditable&gt;5&lt;/span&gt; times.
```

cookiecrook:
Recycling @scottaohara's review with the new diff.

cookiecrook:
> I really appreciate the first one, specifically as a "html doesn't allow this? oh yeah?" it made me grin.

It also keeps the "clickable" functionality of the HTML `<label>` 👍

index.html
```diff
@@ -440,10 +440,11 @@ <h4>Computation steps</h4>
               </ol>
               <div>
                 <summary>Example:</summary>
-                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the user has entered "5" for the embedded <a class="role-reference" href="#textbox">textbox</a>, the complete label is "Flash the screen 5 times", e.g.: </p>
-                <pre class="example highlight">
-                  <code>&lt;div role="checkbox" aria-checked="false"&gt;Flash the screen &lt;span role="textbox" aria-multiline="false"&gt; 5 &lt;/span&gt; times&lt;/div&gt;</code>
-                </pre>
+                <p>Consider a <a class="role-reference" href="#checkbox">check box</a> label that contains a text input field: "Flash the screen [input] times". If the input field has the value "5", the complete label is "Flash the screen 5 times", e.g.:</p>
+                <pre class="example highlight"><code>&lt;label for="flash"&gt;
```
JAWS-test, line 444: With enclosing label the for attribute is unnecessary

scottaohara: It’s not “necessary” but it isn’t invalid and can be necessary for some legacy AT. No need to remove it

scottaohara (approved):

index.html
cookiecrook: As mentioned in my prior comment, removing it would also break the label clickability, and therefore reduce the mainstream usability of the control.

JAWS-test: > As mentioned in my prior comment, removing it would also break the label clickability, and therefore reduce the mainstream usability of the control.

The clickability of the label is also given without the for attribute, if the label encloses the control

cookiecrook: > The clickability of the label is also given without the for attribute, if the label encloses the control

This didn't used to be the case, and old _defensive_ web dev techniques die hard. Somebody, somewhere is still running a browser from a decade ago.

cookiecrook:
@accdc @pkra @spectranaut @jnurthen Please merge?

accdc:
I'm happy merging since we all appear to be in agreement. I'm not sure how to resolve the branch merge conflicts though so will need someone's assistance who has a better handle on Git than I.

cookiecrook:
> I'm not sure how to resolve the branch merge conflicts though

I didn't see any merge conflicts. Merge was successful. Deleting branch.

cookiecrook:
Oh... I see. @jnurthen forward merged main into this one instead of a rebase squash. Thanks.

